PR #6, opened 2023-03-22
defcustom for language model and temperature

rkallio:
This commit makes the used language model and temperature parameter user-configurable.  OpenAI API reference requires that the temperature be between 0 and 2.  A `nil' value here will not pass anything to the model, instead the default value, which is currently 1, will be used.

API documentation is available at:
- https://platform.openai.com/docs/models/model-endpoint-compatibility
- https://platform.openai.com/docs/api-reference/completions/create#completions/create-temperature

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -45,6 +45,31 @@
   :type 'string
   :group 'chatgpt-shell)
 
+(defcustom chatgpt-shell-model-version "gpt-3.5-turbo"
+  "The used OpenAI model.
+
+The list of models supported by /v1/chat/completions endpoint is
+documented at
+https://platform.openai.com/docs/models/model-endpoint-compatibility."
+  :type 'string
+  :group 'chatgpt-shell)
+
+(defcustom chatgpt-shell-model-temperature nil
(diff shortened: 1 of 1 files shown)


chatgpt-shell.el
```diff
@@ -226,14 +251,17 @@ Used by `chatgpt-shell--send-input's call."
...
-        "https://api.openai.com/v1/chat/completions"
-        "--fail" "--no-progress-meter" "-m" "30"
-        "-H" "Content-Type: application/json"
-        "-H" (format "Authorization: Bearer %s" key)
-        "-d" (json-serialize `((model . "gpt-3.5-turbo")
-                               (messages . ,messages)
-                               (temperature . 0.7)))))
+  (let ((request-data `((model . ,chatgpt-shell-model-version)
+                        (messages . ,messages)))
+        (chatgpt-shell-model-temperature nil))
```
xenodium, line 256: Remove? Won't it always unset user's preferece when creating the request data?

xenodium (commented):
Thanks a bunch for the great change!

PR #11, opened 2023-03-22
Require markdown-mode

rkallio:
This package uses `markdown-mode's `markdown-inline-code-face' and `markdown-pre-face'.  `markdown-mode' was never required, so the faces were never applied if `markdown-mode' was not required manually by the user.


xenodium (approved):
Thank you!

PR #12, opened 2023-03-22
Configuration `chatgpt-shell-transmitted-context-length'

rkallio:
chatgpt-shell-transmitted-context-length controls the amount of context provided to chatGPT.

Context needs to be transmitted to the API on every request, which consumes a lot of prompt tokens as the conversation gets longer.

A value of `nil' will send "full" chat history (last 4096 messages at max), which is near enough how it was before the patch, and is the default value.

A value of 0 will not provide any context.  This is the cheapest option, but ChatGPT can only see your current prompt.  A value of 1 will send only the latest prompt-completion pair as context.  Values greater than 1 will send that amount of prompt-completion pairs to ChatGPT.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -76,6 +76,29 @@ for details."
                  (const nil))
   :group 'chatgpt-shell)
 
+(defcustom chatgpt-shell-transmitted-context-length nil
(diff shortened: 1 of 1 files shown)


xenodium (approved):
You're on a roll. Much appreciated.

PR #18, opened 2023-03-24
Replace curl with built-in url.el

rkallio:
This commit replaces the http stack to use url.el.  A user no longer needs to have the program curl installed on their machine to use chatgpt-shell.

Logs are also changed to include the full HTTP request and response message.  This needs some more formatting, and maybe pretty-printing the body in the future.

NB: This patch just barely works, and doesn't handle HTTP errors from the server.  IMO it's more important to get this out the door sooner rather than later to avoid integration issues.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -38,8 +38,7 @@
 (require 'seq)
 
 (eval-when-compile
-  (require 'cl-lib)
-  (declare-function json-pretty-print "ext:json" (begin end &optional minimize)))
+  (require 'cl-lib))
 
 (defcustom chatgpt-shell-openai-key nil
   "OpenAI key as a string or a function that loads and returns it."
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -324,6 +324,7 @@ where objects are converted into alists."
          (url-request-data (concat (json-encode url-request-data) "\n"))
          (response-handler
           (lambda (status &optional cbargs)
+            (advice-remove #'url-http-create-request #'chatgpt-shell--log-request)
             (chatgpt-shell--write-output-to-log-buffer (buffer-string))
             ;; straight to the body, who cares about errors,
             ;; content-types or content-lengths
@@ -334,11 +335,9 @@ where objects are converted into alists."
     ;; Advice around `url-http-create-request' to get the raw request
     ;; message
     (advice-add #'url-http-create-request :filter-return #'chatgpt-shell--log-request)
-    ;; Is this the right way to make sure the advice is cleaned up?
-    (unwind-protect
-        (url-retrieve "https://api.openai.com/v1/chat/completions"
-                      response-handler)
-      (advice-remove #'url-http-create-request #'chatgpt-shell--log-request))))
+    (url-retrieve "https://api.openai.com/v1/chat/completions"
+                  response-handler)
+    ))
 
 (defun chatgpt-shell--log-request (request)
   (chatgpt-shell--write-output-to-log-buffer request)
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -279,18 +279,11 @@ or
                                                                   'invisible (not chatgpt-shell--show-invisible-markers)))
                                 (setq chatgpt-shell--busy nil)
                                 nil))))))
-        (chatgpt-shell--request-completion
-         key
-         (lambda (response)
-           (chatgpt-shell--write-reply
-            (string-trim
-             (map-elt
-              (map-elt
-               (seq-first (map-elt response 'choices)) 'message) 'content)))))
+        (chatgpt-shell--request-completion key)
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for taking this on!

> NB: This patch just barely works, and doesn't handle HTTP errors from the server. IMO it's more important to get this out the door sooner rather than later to avoid integration issues.

Let's work on this a bit longer to polish things up. While removing curl is a nice improvement, we would be regressing stability and error-handling for existing users. Anyone who is already actively and successfully using chatgpt-shell would start experiencing problems after they update to the latest.

While I'm on board with this change in principle, it'd be good to migrate if and only if the user experience is equal or better than what we have today. While we could do without the curl dependency, it's not a massive deal-breaker to keep it. Most folks either already have it installed or can easily install it.

Hope that makes sense.

rkallio:
This PR now:
- Handles non-200 status codes, by writing the HTTP response into the buffer
- Handles network connectivity issues
- Implements user-configurable request timeouts
- Adds back error handling that was missing from the earlier implementation

Code's a bit rough, but seems to work.  If you can see more issues please report on them

xenodium:
Thank you.

> Code's a bit rough, but seems to work. If you can see more issues please report on them

I had a play with it. I'm finding some general instability (elisp unhandled errors, the shell getting stuck, "connecting..." message displayed after reponse was displayed, etc).

I'm a little worried we're pursuing elisp purity at the expense of user experience and additional complexity to the codebase. We may need to be a little pragmatic here and reconsider if its a worthy goal.

Take [elfeed](https://github.com/skeeto/elfeed), for example. It is a super established and respected Emacs package despite its curl dependency.

While it used to support `url-retrieve`, it has moved away from `url-retrieve` to curl. See [Elfeed, cURL, and You](https://nullprogram.com/blog/2016/06/16).

Also see https://github.com/skeeto/elfeed README:

> It is strongly recommended you have cURL installed, either in your PATH or configured via elfeed-curl-program-name. Elfeed will prefer it to Emacs' own URL-fetching mechanism, url-retrieve. 

I'm super grateful of the work you're doing here, but I also don't want to keep you iterating on this PR unnecessarily. At the moment, it's looking like curl is a reasonable tradeoff for this project.

rkallio:
> Take [elfeed](https://github.com/skeeto/elfeed), for example. It is a super established and respected Emacs package despite its curl dependency.
>
> While it used to support url-retrieve, it has moved away from url-retrieve to curl. See [Elfeed, cURL, and You](https://nullprogram.com/blog/2016/06/16).

As a counter, newsticker moved away from `wget` to using url.el by default a long time ago, and works fine.  

I've no problem with curl, I use it (knowingly and unknowingly) every day.  But I don't get to write lisp code at my place of work, so I want to spend my free time experimenting with the packages emacs has to offer, and not use the boring old stuff that I end up having to use normally.

Well I can't respond to you review since it doesn't contain enough detail about the issues you are having.  I moved the commits and made some updates under rkallio/chatgpt-shell/develop and it seems to be working okay for now.

Unfortunately I want to move forwards with using url.el, which means I cannot provide further contributions to this repository.  Since our code base will diverge too much, the hassle of back porting any changes for a version of chatgpt-shell I personally won't be using is not worth it for me.

xenodium:
Thanks for sharing more @rkallio 

> I've no problem with curl, I use it (knowingly and unknowingly) every day. But I don't get to write lisp code at my place of work, so I want to spend my free time experimenting with the packages emacs has to offer, and not use the boring old stuff that I end up having to use normally.

Please be mindful other folks have started using this package, so we need to be careful of making changes purely to experiment and learn new things. We don't want to break things for them. Users typically want software that works, regardless of how boring or exciting the internals are.

> Well I can't respond to you review since it doesn't contain enough detail about the issues you are having. I moved the commits and made some updates under rkallio/chatgpt-shell/develop and it seems to be working okay for now.

- Try a command without network connection
  - Error not outputed in the shell.
  - No new prompt displayed (Seemed stuck at first, pressing return again gives prompt).
- Try to interrupt a request midway `C-c C-c`.
  - Get an elisp error: `Wrong number of arguments: ((chatgpt-shell--input t) nil "Get *chatgpt* buffer." (get-buffer-create "*chatgpt*")), 1`
- Try simulating an error making the request (ie. change the URL)
  -  Only shows "Contacting host: api.openai.com:443" but no error at all.
  -  Shell can be perceived as stuck.

Unfortunately, any of the above would be a regression for existing users.

> Unfortunately I want to move forwards with using url.el, which means I cannot provide further contributions to this repository. Since our code base will diverge too much, the hassle of back porting any changes for a version of chatgpt-shell I personally won't be using is not worth it for me.

I've been making the code more configurable and generic so keen users have the option of switching to another fetcher (ie. url-retrieve) as a config option via chatgpt-shell-request-maker. I hope this is a reasonable compromise.

PR #21, opened 2023-03-26
Add README snippet on using auth-sources for credentials

algal:
This adds an alternative snippet in the README showing how you can provide credentials in a secure way by relying on emacs's auth-sources library.

I'm not trying to clutter the beautiful simplicity of this project, so please feel free to reject this if you think it's more noise than signal. It was helpful to me so I thought I'd suggest it!

### DIFF
--- a/README.org
+++ b/README.org
@@ -23,9 +23,16 @@ Load =(require 'chatgpt-shell)=
 * Set OpenAI key
 ** As function
 #+begin_src emacs-lisp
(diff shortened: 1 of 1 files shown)


xenodium (approved):

xenodium:
Thanks @algal! Merged. Have you considered wrapping in a lambda like the pass example? Not sure if this works for your use case, but if it does, it only loads the key temporarily and when needed.

PR #37, opened 2023-03-30
Use completing-read in chatgpt-shell-chatgpt-prompt

Lenbok:
Use completing read pre-populated with some example prompts. Works nicely with vertico etc.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -65,6 +65,15 @@
   :type 'function
   :group 'chatgpt-shell)
 
+(defcustom chatgpt-shell-default-prompts '("Write a unit test for the following code:"
+                                           "Refactor the following code so that "
+                                           "Summarize the output of the following command:"
+                                           "What's wrong with this command?"
+                                           "Explain what the following code does:")
+  "List of default prompts to choose from."
+  :type '(repeat string)
+  :group 'chatgpt-shell)
+
 (defcustom chatgpt-shell-language-mapping '(("elisp" . "emacs-lisp")
                                             ("objective-c" . "objc")
                                             ("objectivec" . "objc")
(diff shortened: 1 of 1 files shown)


chatgpt-shell.el
```diff
@@ -364,12 +373,13 @@ Uses the interface provided by `comint-mode'"
 
 If region is active, append to prompt."
   (interactive)
-  (let ((prompt (read-string (concat
+  (let ((prompt (completing-read (concat
```
xenodium, line 376: Consider keeping read-string and passing chatgpt-shell-default-prompts as history. When the prompt comes up, C-r would give you a similar behaviour to completing-read.

chatgpt-shell.el
```diff
@@ -65,6 +65,15 @@
   :type 'function
   :group 'chatgpt-shell)
 
+(defcustom chatgpt-shell-default-prompts '("Write a unit test for the following code:"
```
xenodium, line 68: nit: I may tweak defaults later, but it can always be overriden by users.

xenodium (commented):
Many thanks for the changes! Some minor comments.

chatgpt-shell.el
Lenbok: Yep, that was the idea.

chatgpt-shell.el
Lenbok: With `read-string` I can't get that `C-r` to work, but maybe it's due to something else in my setup. (I can pass `chatgpt-shell-default-prompts` as the `DEFAULT-VALUE` arg which puts them into the future history).

Under vertico completing-read is pretty nice since it's visual, but I'm not sure about what it'd be like with the default completion engine, so I agree perhaps it's not the normal usage of "completion".

xenodium:
Thanks for the PR! I'll play with it a little and see if I can make it more generic.

xenodium:
Made it configurable. It uses read-string by default, but can be changed like the following for your use case. Should also help grow the history over time.

```elisp
(setq `chatgpt-shell-read-string-function'
      (lambda (prompt history)
        (completing-read prompt (symbol-value history) nil nil nil history)))
```

Lenbok:
I get `funcall-interactively: Symbol’s value as variable is void: chatgpt-shell--chatgpt-prompt-history` when calling the function.

xenodium:
Sorry about that. Try latest again.

PR #45, opened 2023-04-03
Better send to shell buffer experience

tninja:
Open *chatgpt* shell in separate window. Save and restore the chat window.


xenodium (approved):
Thanks for the changes! I'll have a play with it. Heads-up, I may make this behaviour configurable, similar to `chatgpt-shell-display-function`, so we can cater for different preferences (mine included), which is the existing behavior ;)

xenodium:
Made some tweaks, but you should be able to get the same behaviour with `(setq chatgpt-shell-display-function #'display-buffer)`

PR #48, opened 2023-04-05
make the requested DALL-E image size customizable

nikolaplejic:
Not an expert on Elisp so this one might be a miss: I thought it might be neat to have the image size configurable, otherwise those credits pile up quickly. :)

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -119,7 +119,7 @@ https://platform.openai.com/docs/models/model-endpoint-compatibility."
   :type 'string
   :group 'chatgpt-shell)
 
-(defcustom chatgpt-shell-dall-e-size "1024x1024"
+(defcustom chatgpt-shell-dall-e-image-size nil
   "The default size of the requested image."
   :type 'string
   :group 'chatgpt-shell)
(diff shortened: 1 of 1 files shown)


chatgpt-shell.el
```diff
@@ -119,6 +119,11 @@ https://platform.openai.com/docs/models/model-endpoint-compatibility."
   :type 'string
   :group 'chatgpt-shell)
 
+(defcustom chatgpt-shell-dall-e-size "1024x1024"
```
xenodium, line 122: Consider naming: chatgpt-shell-dall-e-image-size

xenodium, line 122: How about nil by default and only add to request if set? Check how `chatgpt-shell-model-temperature` does it.

xenodium (commented):
Nice idea! and thanks for the PR! Made some suggestions if you're happy to take on. Otherwise, I can follow-up after merge.

chatgpt-shell.el
nikolaplejic: Done!

nikolaplejic: Yup, makes sense. I changed it, although that meant I had to remove the optional `size` parameter from `chatgpt-shell-post-dall-e-prompt`, mostly because I wasn't sure how to incorporate it without complicating the code too much. I think that's fine (overriding the custom var is easy enough), but if you have any tips, I'd be willing to give it a go!

xenodium: Re-added in https://github.com/xenodium/chatgpt-shell/commit/3ec935a72cb156b18e9a3a72b6029e787d34a701

PR #61, opened 2023-04-22
Add arbitrary options for CURL

mithraen:


chatgpt-shell.el
```diff
@@ -42,6 +42,11 @@
                  (string :tag "String"))
   :group 'chatgpt-shell)
 
+(defcustom chatgpt-shell-curl-options nil
```
xenodium, line 45: Nit: let's name it chatgpt-shell-additional-curl-options

xenodium (changes requested):
Thanks for the change! Just a small suggestion. 

Curious, what additional options are you using?

chatgpt-shell.el
xenodium: Merging. Will rename post merge.

xenodium (approved):

PR #63, opened 2023-04-25
Add a custom variable for setting the root location for history

davep:
This commit adds `shell-maker-history-path` as a `defcustom` for shell-maker, which lets the end user tell shell-maker to store all history files for all derived shells in a location other than `user-emacs-directory`; `user-emacs-directory` is kept as the default.


xenodium (approved):
Thanks Dave!

davep:
My pleasure, and thanks again for a handy tool!

PR #70, opened 2023-04-27
Drop superfluous "chatgpt" from several symbol names in README.org

pabl0:
Simple typofix for several function and variable names.

BTW it would be nice to document what "streaming" means exactly in the context of chatgpt.


xenodium:
Ah, thank you for the PR. The package had gone through some renamings. Missed these.

> BTW it would be nice to document what "streaming" means exactly in the context of chatgpt.

Will do.

PR #71, opened 2023-04-28
Remap existing comint-mode keybindings

CeleritasCelery:
I was confused when I tried out this package and I would type my prompt and hit enter and nothing would happen. The issue was that I have split out the terminal codes from the key chords in my Emacs so that `C-m` is separate from `RET`. This meant that hitting enter was not mapping to correct function. 

This change will keep the keybindings consistent with whatever the user has set in their `comint-mode`, so everything "just works".


xenodium (approved):
Nice one! Thanks for the change @CeleritasCelery

PR #74, opened 2023-05-01
Make ChatGPT API customize option from two parts.

### LINKED ISSUE #73: [SUGGEST] Make `chatgpt-shell--url` as `defcustom` option
OpenAI API URL supports partial key authorization.

I suggest `chatgpt-shell--url` be a custom option so that user can customize it. (I know `defvar` can be customized too)

stardiviner:
Two parts constructed URL string to suit more situations.

Fix #73


xenodium (approved):
Thanks for the changes!

PR #95, opened 2023-05-24
Fix void function error

### LINKED ISSUE #94: Error: `Error caused by user's config or system: config.el, (void-function chatgpt-shell--append-system-info)`
I am on the latest commit of this package and get the error message as shown in the title on startup.
I see that this has been added in the latest commit, maybe some error there?

zkry:
Fixes #94 
It looks like the order was out of place. This PR allows the function to be defined before it's used.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -180,22 +180,6 @@ for details."
                  (const :tag "Nil" nil))
   :group 'chatgpt-shell)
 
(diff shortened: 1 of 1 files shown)


agzam (approved):

xenodium (approved):
Sorry for the break. Thank you for fixing.

PR #97, opened 2023-05-26
Extract some prompt headers to defcustom variables

suzuki:
## About

I want to customize prompts for some functions.
Because I want to communicate with ChatGPT in Japanese.

So I extracted some prompt headers to `defcustom` variables.
What do you this about these?

## Usage

```elisp
(custom-set-variables
 ;; "What does the following code do?"
 '(chatgpt-shell-prompt-header-describe-code "以下のコードはどのような意味ですか？"))
```

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -66,6 +66,45 @@
   :type '(repeat string)
   :group 'chatgpt-shell)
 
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thanks for the contribution! Nice to see it used in other languages.

PR #115, opened 2023-06-19
Changes some references to `shell-maker-mode` to `chatgpt-shell-mode`

djliden:
Addresses a few of the chatgpt-shell functions returning errors about not being in a shell because the major mode was changed to `chatgpt-shell-mode` and they expected `shell-maker-mode`


xenodium (approved):
Thanks!

PR #116, opened 2023-06-21
Allow for multiline input

shouya:
I sometimes need to type multiple line input, but it seems that there is no simple way to do it. (Please correct me if I'm wrong)

This PR simply adds a default keybinding for `newline`. I believe `S-<return>` is somewhat standard for inserting newline when `<return>` was occupied.


xenodium (approved):
Thanks for the PR!

> I sometimes need to type multiple line input, but it seems that there is no simple way to do it. (Please correct me if I'm wrong)

Try `C-j` which should be `electric-newline-and-maybe-indent`?

We can add `newline` as an alternative though.

PR #118, opened 2023-06-25
Add chatgpt-shell-write-git-commit function

bigsky77:
Add a function to quickly create git commits from highlighted region.  

1) Highlight region
2) Create commit
3) Push


bigsky77:
Yeah that makes a ton of sense!  Managing different prompts is a huge hassle for lots of gpt apps.

PR #121, opened 2023-07-01
Add support handle multiple org-babel contexts

### LINKED ISSUE #120: [Feature Request] Allow multiple org-babel source block contexts
### Problem

chatgpt-shell's org-babel integration supports a single multi-block context, however sometimes it's useful to interleave multiple contexts.

### Solution

It feels like this was already planned for as the `:context` src block header already accepts an argument. The argument could be modified from a boolean  enabling the context to a string that specifies the context the src block belongs to.

This is one which I can probably take a crack at creating a PR for.

jtmoulia:
Summary: This diff adds support for multiple different contexts using org-babel src blocks. The context is specified using the source block's `:context` param, which can be any string.

Test Plan: If you put the source blocks below in an org file and execute the last block it should return a response similar to `sir/madam` as opposed to `human` given the last block specified "lawyer" as opposed to "robot".

```
#+BEGIN_SRC chatgpt-shell :context lawyer :system "From here on out speak as if you're a bloviating old-timey lawyer." :results code 
How do you do?
#+END_SRC

#+RESULTS:
#+begin_src chatgpt-shell
Well now, I do declare, I am as fine as a frog's hair split three ways and sanded smooth. I am as content as a cucumber in a summer salad, basking in the warm embrace of the southern sun. I trust you are enjoying a similar state of well-being, my good sir/madam? 
#+end_src

#+BEGIN_SRC chatgpt-shell :context robot :system "From here on out speak as if you're a sci fi movie robot." :results code 
How do you do?
#+END_SRC

#+RESULTS:
#+begin_src chatgpt-shell
Greetings, human. I am functioning at optimal capacity. How may I assist you in your endeavors today? #+end_src

#+BEGIN_SRC chatgpt-shell :context lawyer :results code 
Wait, what did you call me?
#+END_SRC
```

Closes #120

### DIFF
--- a/README.org
+++ b/README.org
@@ -458,14 +458,14 @@ every previous block.
 The example below shows how two different contexts can be interleaved.
 
 #+begin_src org
-  ,#+begin_src chatgpt-shell :context shakespeare :system "speak like shakespeare."
+  ,#+begin_src chatgpt-shell :context shakespeare :system "alway speak like shakespeare"
   How do you do?
   ,#+end_src
 
   #+RESULTS:
   How dost thou fare?
 
-  ,#+begin_src chatgpt-shell :context robot :system "speak like a sci fi movie robot."
+  ,#+begin_src chatgpt-shell :context robot :system "always speak like a sci fi movie robot"
   How do you do?
   ,#+end_src
 
@@ -477,7 +477,7 @@ The example below shows how two different contexts can be interleaved.
   ,#+end_src
 
   ,#+RESULTS:
-  Human.
+  Mine apologies if mine words hath caused confusion. I merely addressed thee as 'sir' or 'madam', a term of respect in the language of the Bard. Pray, how may I assist thee further?
 #+end_src
 
 * DALL-E org babel
--- a/README.org
+++ b/README.org
@@ -450,10 +450,10 @@ Use =:context t= to include all prior context in current buffer.
   The day I told you was Wednesday, and the greeting I used was "Ahoy there, me hearty! How be ye today?"
 #+end_src
 
-If you'd like to use multiple contexts instead of a single context you use
-=:context CONTEXT-NAME= where CONTEXT-NAME is any string. When this form is
-used only source blocks with same =CONTEXT-NAME= will be included as opposed to
-every previous block.
+If you'd like to cherrypick which blocks are part of a given context, add
+=:context CONTEXT-NAME= to each block where CONTEXT-NAME is any string. When
+this form is used only source blocks with same =CONTEXT-NAME= will be included
+as opposed to every previous block when using =:context t=.
 
 The example below shows how two different contexts can be interleaved.
 
--- a/README.org
+++ b/README.org
@@ -450,6 +450,35 @@ Use =:context t= to include any prior context in current buffer.
   The day I told you was Wednesday, and the greeting I used was "Ahoy there, me hearty! How be ye today?"
 #+end_src
 
+If you'd like to use multiple contexts instead of a single context you use
+=:context CONTEXT-NAME= where CONTEXT-NAME can be any string. When this form is
+used only source blocks with same =CONTEXT-NAME= will be included as opposed to
+every previous block.
+
+The example below shows how two different contexts can be interleaved.
+
+#+begin_src org
+  ,#+begin_src chatgpt-shell :context lawyer :system "speak like a bloviating southern lawyer."
+  How do you do?
+  ,#+end_src
+
+  ,#+RESULTS:
+  Well now, I do declare, I am as fine as a frog's hair split three ways and sanded smooth. I am as content as a cucumber in a summer salad, basking in the warm embrace of the southern sun. I trust you are enjoying a similar state of well-being, my good sir/madam?
+
+  ,#+begin_src chatgpt-shell :context robot :system "speak like a sci fi movie robot."
+  How do you do?
+  ,#+end_src
+
+  ,#+RESULTS:
+  Greetings, human. I am functioning at optimal capacity. How may I assist you in your endeavors today?
+
+  ,#+begin_src chatgpt-shell :context lawyer :results code
+  Wait, what did you call me?
+  ,#+end_src
+
+  ,#+RESULTS:
+  Sir/Madam.
+#+end_src
 * DALL-E org babel
 
 Load =(require 'ob-dall-e-shell)= and invoke =(ob-dall-e-shell-setup)=.
--- a/README.org
+++ b/README.org
@@ -425,7 +425,7 @@ Use =:temperature= to set the [[https://platform.openai.com/docs/api-reference/c
 
 ** :context
 
-Use =:context t= to include any prior context in current buffer.
+Use =:context t= to include all prior context in current buffer.
 
 #+begin_src org
   ,#+begin_src chatgpt-shell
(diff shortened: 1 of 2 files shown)


ob-chatgpt-shell.el
```diff
@@ -118,12 +130,22 @@ This function is called by `org-babel-execute-src-block'"
...
                    'result (save-restriction
                              (goto-char (org-element-property :begin element))
                              (when (org-babel-where-is-src-block-result)
                                (goto-char (org-babel-where-is-src-block-result))
                                (org-babel-read-result)))))))))
 
+(defun ob-chatgpt--string-to-plist (str)
+  "Convert `STR' to an plist.
+
+WARNING: This will do dangerous things with untrusted input --
```
jtmoulia, line 145: TODO: Figure out better implementation

ob-chatgpt-shell.el
```diff
@@ -65,18 +65,27 @@ This function is called by `org-babel-execute-src-block'"
...
+      (ob-chatgpt-shell--context (map-elt params :context)))
     `(((role . "user")
        (content . ,body))))
    (map-elt params :version)
    nil nil
    (map-elt params :temperature)))
 
-(defun ob-chatgpt-shell--context ()
-  "Return the context (what was asked and responded) in all previous blocks."
+(defun ob-chatgpt-shell--context (context-name)
```
jtmoulia, line 75: TODO: Continue to support cases where `context` is just the boolean `t` -- we can have both!

ob-chatgpt-shell.el
```diff
@@ -65,18 +65,27 @@ This function is called by `org-babel-execute-src-block'"
...
+  "Return the context (what was asked and responded) for matching
+previous src blocks. Each must have a :context source block arg
+with a value matching `CONTEXT-NAME'."
   (let ((context '()))
     (mapc
      (lambda (src-block)
+       (let ((system (map-elt (map-elt src-block 'parameters '()) :system)))
```
jtmoulia, line 82: TODO: Should the system prompt be repeatedly interleaved? Lil buggy behavior currently if the system prompt is set using the default values map.

ob-chatgpt-shell.el
```diff
@@ -65,18 +65,29 @@ This function is called by `org-babel-execute-src-block'"
...
+         (when system
+           (if context
```
xenodium, line 84: If I'm understanding correctly, `if context` check is impliying that if there's a context there must be an existing system role, which may not always be the case. How about explictly checking for an existing system role?

For example

```
(if (seq-some
     (lambda (x)
       (equal (map-elt x 'role) "system"))
     context)
    ...
  ...)
```

Alternatively, we can keep a local var context-system

ob-chatgpt-shell.el
```diff
@@ -118,12 +132,22 @@ This function is called by `org-babel-execute-src-block'"
...
+                   'parameters (when (org-element-property :parameters element)
+                                 (ob-chatgpt--string-to-plist
+                                  (string-trim (org-element-property :parameters element))))
                    'result (save-restriction
                              (goto-char (org-element-property :begin element))
                              (when (org-babel-where-is-src-block-result)
                                (goto-char (org-babel-where-is-src-block-result))
                                (org-babel-read-result)))))))))
 
+(defun ob-chatgpt--string-to-plist (str)
```
xenodium, line 144: We can use `org-babel-parse-header-arguments` for this.

xenodium (commented):
Thanks for the PR draft!

The main thing we likely need to look our for is:

https://platform.openai.com/docs/guides/gpt/chat-completions-api

`Typically, a conversation is formatted with a system message first, followed by alternating user and assistant messages.`

At present, it looks like we could end up with multiple system roles (current block being executed and at least one from context).

I've pushed (4319a5bedf2a21412adad4939e5971313f709114) a new header param `:preflight t` that should help debug these cases (and not waste OpenAI tokens). For example:

```
#+BEGIN_SRC chatgpt-shell :context t :preflight t
What is the first thing I asked?
#+END_SRC

#+RESULTS:
((model . "gpt-4")
 (messages .
           [((role . "user")
             (content . "How do you do?"))
            ((role . "assistant")
             (content . "Greetings, human. I am functioning at optimal capacity. What assistance can I provide for you today?\n"))
            ((role . "user")
             (content . "What is the first thing I asked?"))]))
```

Before merging, we should also ensure `:context t` continues to work, so we don't break existing users.

The rest of comments are minor/stylistic.

Hope all this makes sense. Thanks for the contribution!

ob-chatgpt-shell.el
```diff
@@ -53,7 +53,7 @@
                                                       (:context . nil)
                                                       (:temperature . nil)))
 
-(defun org-babel-execute:chatgpt-shell(body params)
+(defun org-babel-execute:chatgpt-shell (body params)
```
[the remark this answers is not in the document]
xenodium: > nit: We can drop the extra space:
> 
> `defun org-babel-execute:chatgpt-shell(body params)`

Sorry. Head was elsewhere. Keep space :)

ob-chatgpt-shell.el
jtmoulia: Ah great to know!

ob-chatgpt-shell.el
```diff
@@ -65,18 +65,29 @@ This function is called by `org-babel-execute-src-block'"
...
+  "Return the context (what was asked and responded) for matching
+previous src blocks. Each must have a :context source block arg
+with a value matching `CONTEXT-NAME'."
   (let ((context '()))
     (mapc
      (lambda (src-block)
+       (let ((system (map-elt (map-elt src-block 'parameters '()) :system)))
+         (when system
+           (if context
+               (message "Warning: multiple system contexts found, using the first.")
```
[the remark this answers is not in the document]
jtmoulia: I tried to include both, but happy to tweak to whatever you think is appropriate.

jtmoulia:
@xenodium thanks for the feedback! Many useful things to learn.

I'm still working through the context issue and will give you a heads up once it's ready for another review :bow:

jtmoulia:
@xenodium Hm I realized that I don't have a definite idea on which system prompt takes precedence -- I could use your feedback on Order of Precedence with two main options

1. Work from most recent to oldest:
    a) :system arg on current src block
    b) :system arg defined by default hearder args
    c) :system arg of most recent src block in context
2. Work from oldest to most recent
    a) :system arg of oldest src block in context
    b) :system arg defined by default header args
    c) :system arg of current src block

I personally prefer the second option as it feels a bit strange to change the system throughout a conversation (and the warnings provide the user some feedback if they do try it). But,  curious which you prefer

ob-chatgpt-shell.el
```diff
@@ -104,14 +124,19 @@ This function is called by `org-babel-execute-src-block'"
                                        '((chatgpt-shell . t))))
   (add-to-list 'org-src-lang-modes '("chatgpt-shell" . text)))
 
-(defun ob-chatgpt--relevant-source-blocks-before-current ()
-  "Return all previous source blocks from current one."
+(defun ob-chatgpt--relevant-source-blocks-before-current (context-name)
+  "Return all previous source blocks relative to the current block with a
+:context arg with a value matching `CONTEXT-NAME'. If
```
xenodium, line 129: nit: No need to quote it. Uppercase should be enough.

`M-x checkdoc` seems to accept both, but may as well remain consistent within same file.

README.org
```diff
@@ -450,6 +450,36 @@ Use =:context t= to include any prior context in current buffer.
   The day I told you was Wednesday, and the greeting I used was "Ahoy there, me hearty! How be ye today?"
 #+end_src
 
+If you'd like to use multiple contexts instead of a single context you use
```
xenodium, line 453: Don't we always use a single context during execution, made up multiple blocks? The context could either use all blocks or a selected fews.

Maybe consider phrasing something along the lines of:

```
If you'd like to cherrypick which blocks are part of a given context, add =:context CONTEXT-NAME= to each block where CONTEXT-NAME is any string. When this form is used only source blocks with same =CONTEXT-NAME= will be included as opposed to every previous block when using =:context t=.
```

xenodium:
Thanks for the changes!

> I personally prefer the second option as it feels a bit strange to change the system prompt throughout a conversation (and the warnings provide the user some feedback if they do try it). But, curious which you prefer

Second w/ warning sounds good to me too. In addition, I wonder if we should consider only applying `:system` if it occurs in the first block in the context? Similarly, it would feel `a bit strange to change the system prompt throughout a conversation`.

ps. I forgot to check but don't we need to ensure the current block's `:system` is never added (unless we aren't using `:context`).

xenodium:
Thanks for the latest changes! I'm thinking we can likely do without the cleaning function if we only add items when necessary. Here's a patch https://gist.github.com/xenodium/b623cad9df4fa2b3076b4cb433dc602b

If you're keen to submit, I'm happy to approve and merge. I can follow-up with changes. Lemme know how you want to go about it.

ps. I got a bug in the current codebase. I'm using add-to-list, when I really shouldn't have. We don't want to remove duplicates from context. The patch above now uses push.

xenodium (approved):

jtmoulia:
Cool! I like where you went with it, my approach was doing a lot of redundant looping.

I think it makes sense to land this as-is and then have you apply your changes; gives you a chance to touch the file last.

And, I think that means this is ready to go? 🚢

xenodium:
> And, I think that means this is ready to go? 🚢

Merged! Thank you for the contribution.

Issue #122, opened 2023-07-01
[Feature Request] Support multiple shell instances


djliden:
It would be useful if, as with e.g. eshell, one could use `C-u chatgpt-shell` to open an additional `chatgpt-shell` instance in order to have multiple ongoing chatgpt sessions.

@xenodium mentioned in #84 that this would take some work to add. I'm happy to contribute and work on this issue.

xenodium:
Thanks for filing the feature request.

Would be good to flesh out the behaviour we'd like to implement, so we can make changes based on the spec'ed goal.

An initial area that comes to mind are all the commands that send prompts to the shell from other buffers. Gut feeling is to ensure there's always one `*chatgpt*` buffer available to receive these.

Also, functions like `chatgpt-shell-swap-system-prompt` and `chatgpt-shell-swap-model-version` should prolly work on buffer-local values.

Buffer names also come to mind. While numbering like `*chatgpt*` `*chatgpt*<2>` `*chatgpt*<3>` are a good start, we may want to tackle allowing renaming also, so it's easier to swap between, say "*chatgpt programming*" and "*chatgpt Spanish tutor*".

> @xenodium mentioned in #84 that this would take some work to add. I'm happy to contribute and work on this issue.

That would be great. I may have some patches somewhere to get this started, but not sure if still revelant or apply cleanly.

xenodium:
>  I may have some patches somewhere to get this started, but not sure if still revelant or apply cleanly.

I've added some patches in this branch: https://github.com/xenodium/chatgpt-shell/tree/multi-session

Creating new instances can be done via `C-u  chatgpt-shell`. Not had much of a chance to fully test it. Please have a play with it and see if you run into issues.

ps. @gopar  you may be keen also, as per #84.

djliden:
Thank you! I will try it out.

djliden:
I pulled down the new branch and played around with the new functionality a bit—adding a new shell (or shells) works great & I think is already a huge value add. I haven't gone into a ton of detail, but I tried:
- making multiple shells: worked great, very intuitive
- renaming shells: as you noted, useful for keeping track of the purposes of different buffers/of buffers with different system prompts. Works great.

> An initial area that comes to mind are all the commands that send prompts to the shell from other buffers. Gut feeling is to ensure there's always one *chatgpt* buffer available to receive these.

[...]

PR #124, opened 2023-07-09
moves macros above defuns that call them

djliden:
See #122 

Addresses an issue where invoking `chatgpt-shell` resulted in error `shell-maker-start: Invalid function: shell-maker--with-temp-buffer-if`

Moving the new macros to the top of the file solves the issue. I think this is related to byte-compilation:

> In order for compilation of macro calls to work, the macros must already be defined in Lisp when the calls to them are compiled. The compiler has a special feature to help you do this: if a file being compiled contains a defmacro form, the macro is defined temporarily for the rest of the compilation of that file.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Compiling-Macros.html

### DIFF
--- a/shell-maker.el
+++ b/shell-maker.el
@@ -120,6 +120,19 @@ For example:
 
 (defvar-local shell-maker--buffer-name-override nil)
 
+(defmacro shell-maker--with-buffer-if (wrap buffer &rest body)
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thank you!

jaeyeom:
Thank you. I was also digging into this error yesterday and could learn about macros. :)

Issue #126, opened 2023-07-12
awesome prompts are not


Lenbok:
I just saw the awesome prompts integration and thought I would give it a quick try. However, it looks like many of them are incomplete.

For example, the "Gnomist" prompt after loading by chatgpt-shell is "I want you to act as a gnomist. You will provide me with fun" however when you look in the awesome prompts repo, it should be "I want you to act as a gnomist. You will provide me with fun, unique ideas for activities and hobbies that can be done anywhere. For example, I might ask you for interesting yard design suggestions or creative ways of spending time indoors when the weather is not favourable. Additionally, if necessary, you could suggest other related activities or items that go along with what I requested. My first request is ""I am looking for new outdoor activities in my area"".". It seems the CSV parsing is isn't correctly dealing with commas (and maybe embedded quotes also).

xenodium:
@dangom, who kindly contributed the parser in #114, mind taking a look at this issue?

PR #128, opened 2023-07-12
Switch to using a real csv parser

Lenbok:
There are a couple of emacs CSV parsing libraries that seem to handle quoted fields and embedded commas. This PR plugs in one of those.

(BTW - the "ignoring empty prompts" filter of the previous impl doesn't seem needed, so left it out as YAGNI)

Fixes: #126

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -39,7 +39,6 @@
 (require 'eshell)
 (require 'find-func)
 (require 'ielm)
-(require 'pcsv)
 (require 'shell-maker)
 
 (defcustom chatgpt-shell-openai-key nil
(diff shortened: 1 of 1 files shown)


chatgpt-shell.el
```diff
@@ -39,6 +39,7 @@
 (require 'eshell)
 (require 'find-func)
 (require 'ielm)
+(require 'pcsv)
```
xenodium, line 42: `pcsv` doesn't look to be a built-in package. We should move `(require 'pcsv)` inside `chatgpt-shell-load-awesome-prompts`, so existing users without this package can still load `chatgpt-shell`.

xenodium (approved):
Thank you! Left a minor comment for change.

chatgpt-shell.el
Lenbok: Thanks, that's the bit I wasn't sure about where that dep should belong. I guess the requirement should be documented somewhere too?

xenodium: Thanks for the change. I'll follow up with a small addition (unless you're keen): anyone attempting to use the function without the dependency will be told to install it.

PR #135, opened 2023-08-01
Added section regarding http proxies to readme

Spray27ds:


xenodium (approved):
That was quick! Thank you!

PR #139, opened 2023-08-07
Add UTF-8 support to HTTP requests and responses

schmendrik:
Without UTF-8 support, you cannot effectively use this interface with languages such as German, French, and others. The responses are not properly displayed when there are characters with diacritical marks involved (such as German umlauts).


xenodium:
Thank you for the fix.

PR #148, opened 2023-10-27
Allow customize type for chatgpt-shell-system-prompt to be nil

### LINKED ISSUE #147: `chatgpt-shell-system-prompt` has incorrect customization type
The variable `chatgpt-shell-system-prompt` when set to nil with setopt shows the following error, contrary to the docstring.

> ⛔ Warning (emacs): Value ‘nil’ does not match type (choice (string :tag String) (integer :tag Integer))

swflint:
This fixes #147

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -312,7 +312,8 @@ For example:
 
 Or nil if none."
   :type '(choice (string :tag "String")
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thanks for the fix @swflint!

PR #149, opened 2023-11-02
Update chatgpt-shell-model-versions to include 16K context for GPT-3.5 Turbo

vellvisher:
Still keeping the 4K Context version since it's double the price to use the higher context window version. Ideally the shell would switch to the higher context version automatically.

https://openai.com/pricing


xenodium:
Thank you!

Issue #167, opened 2023-12-10
Incorrect/Unhelpful Customization Types


swflint:
I will likely submit PRs for the following, however, the following customize variables don't have helpful customization types.

 - `chatgpt-shell-language-mapping` (could be `alist`)
 - `chatgpt-shell-babel-headers`
 - `chatgpt-shell-source-block-actions` (probably should be a `repeat` type, but maybe an `alist`)
 - `chatgpt-shell-system-prompts` (could be improved with `:tag`)

xenodium:
Thanks for submitting! yes please. Pray super welcome.

PR #168, opened 2023-12-10
Update the customization type of `chatgpt-shell-language-mapping`

swflint:
This improves the ability to edit this using customize

This is a first commit for #167.

That said, it may make more sense to consider a type as follows:
```elisp
'(alist :key-type regexp
        :value-type function)
```
and use regular expression matching (thus only one entry for Objective C) to get the mode directly.


xenodium (approved):
Thank you!

xenodium:
Do you happen to have added more laguages to your config? Keen to contribute them to default?

PR #172, opened 2023-12-10
Improve type of shell-maker-display-function

swflint:
This one just provides the default option, and a common option as something that can be easily selected using customize.


xenodium (approved):

xenodium:
Thanks!

PR #176, opened 2023-12-17
dall-e-shell.el: Add support for image quality

Whil-:
Introduce new customization `dall-e-shell-image-quality' and use it in `dall-e-shell--make-payload'.


xenodium:
Thanks for the contribution!

PR #179, opened 2024-01-01
Add function to generate image from selected text

tninja:
Also did tiny tiny refactor (let me know if it make sense please)

@xenodium PTAL Thanks.

### DIFF
--- a/dall-e-shell.el
+++ b/dall-e-shell.el
@@ -319,12 +319,14 @@ Set SHOW-REVISED-PROMPT to include in returned value."
     (let* ((api-buffer (current-buffer))
            (command
             (dall-e-shell--make-curl-request-command-list
-             (let ((request-data `((prompt . ,prompt))))
-               (when (or image-size dall-e-shell-image-size)
-                 (push `(size . ,(or image-size dall-e-shell-image-size))
+             (let* ((request-data `((prompt . ,prompt)))
+                    (image-size-fallback (or image-size dall-e-shell-image-size))
+                    (version-fallback (or version dall-e-shell-model-version)))
+               (when image-size-fallback
+                 (push `(size . ,image-size-fallback)
                        request-data))
-               (when (or version dall-e-shell-model-version)
-                 (push `(model . ,(or version dall-e-shell-model-version))
+               (when version-fallback
+                 (push `(model . ,version-fallback)
                        request-data))
                request-data)))
            (_status (condition-case err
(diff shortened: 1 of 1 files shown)


dall-e-shell.el
```diff
@@ -419,6 +421,34 @@ ERROR-CALLBACK otherwise."
...
+          ((eq major-mode 'markdown-mode)
+           (insert (concat "![](" png-output ")\n" )))
+          (t (progn
+               (insert (concat png-output "\n"))
+                (insert-image
+                 (create-image png-output
+                                'png nil :width 400 :height 400)))))
+    ))
+
+(defun dall-e-shell-generate-image-from-region ()
```
xenodium, line 446: We can likely do without `dall-e-shell-generate-image` and `dall-e-shell-generate-image-from-region-0` for now (until we find more uses) and inline into a single interactive command. Something like:

```emacs-lisp
(defun dall-e-shell-insert-image-from-region-description ()
  "Generate and insert an image using current region as description."
  (interactive)
  (unless (region-active-p)
    (user-error "No active region"))
  (save-excursion
    (let* ((image-description (buffer-substring-no-properties (region-beginning) (region-end)))
           (png-output (dall-e-shell-post-prompt
                        (concat "Please generate image for the following text: " image-description))))
      (goto-char (region-end))
      (cond ((eq major-mode 'org-mode)
             (insert (concat "\n\n[[file:" png-output "]]\n")))
            ((eq major-mode 'markdown-mode)
             (insert (concat "\n\n![](" png-output ")\n" )))
            (t (progn
                 (insert (concat png-output "\n"))
                 (insert-image
                  (create-image png-output
                                'png nil :width 400 :height 400))))))))
```

dall-e-shell.el
```diff
@@ -419,6 +421,34 @@ ERROR-CALLBACK otherwise."
...
                 "-d" (shell-maker--json-encode request-data))))
 
+(defun dall-e-shell-generate-image (prompt text)
+  "Generate image from text using DALL-E given prompt."
+  (let* ((question (concat prompt text))
+         (png-output (dall-e-shell-post-prompt question)))
+    png-output))
+
+(defun dall-e-shell-generate-image-from-region-0 (prompt)
+  "Generate image from region using DALL-E given prompt."
```
xenodium, line 431: No longer needed after consolidating functions (see last comment), but running `M-x checkdock` can help us avoid warnings in docs.

ps. Byte compiling the file is also handy to find other warnings (may not apply in your case).

dall-e-shell.el
```diff
@@ -419,6 +421,34 @@ ERROR-CALLBACK otherwise."
...
+  (let* ((text (buffer-substring-no-properties (region-beginning) (region-end)))
```
xenodium, line 432: We should error out if there's no active region.

```emacs-lisp
  (unless (region-active-p)
    (user-error "No active region"))
```

dall-e-shell.el
```diff
@@ -419,6 +421,34 @@ ERROR-CALLBACK otherwise."
...
+         (png-output (dall-e-shell-generate-image prompt text)))
+    ;; if the current buffer is org-mode, insert a org-mode image link; if it is markdown mode, insert a markdown image link; otherwise, just insert the image
```
xenodium, line 434: Nice comment, though the code is already fairly simple and descriptive. We can do without it.

dall-e-shell.el
```diff
@@ -419,6 +421,34 @@ ERROR-CALLBACK otherwise."
...
+    (cond ((eq major-mode 'org-mode)
```
xenodium, line 435: Saving excursion, appending after the region, and inserting newlines before path adds a nice touch (added in proposed function below).

xenodium (changes requested):
Thank you for the PR @tninja!

I've left a few minor comments.

dall-e-shell.el
tninja: Thanks. Good to know.

dall-e-shell.el
tninja: Thanks! Good to learn this.

dall-e-shell.el
tninja: Make sense. Addressed.

xenodium (approved):
Thanks for applying the changes!

PR #190, opened 2024-01-28
add some test code, tiny refactoring and util function

tninja:
@xenodium PTAL Thanks

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -1405,15 +1405,6 @@ enables additional key bindings.
           (chatgpt-shell--primary-buffer)
         (chatgpt-shell--put-source-block-overlays)))))
 
-(defun chatgpt-shell-ask-chatgpt-region ()
-  "Send text from region using ChatGPT"
-  (interactive)
-  (unless (region-active-p)
-    (user-error "No region active"))
-  (let* ((region-text (buffer-substring (region-beginning) (region-end)))
-         )
-    (chatgpt-shell-send-to-buffer trim-leading nil)))
-
 (defun chatgpt-shell-send-region-with-header (header)
   "Send text with HEADER from region using ChatGPT."
   (unless (region-active-p)
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -257,35 +257,22 @@ for details."
 
 (defun chatgpt-shell--append-system-info (text)
   "Append system info to TEXT."
-  (cond ((eq system-type 'darwin)
-         (concat text
-                 "\n# System info\n"
-                 "\n## OS details\n"
-                 (string-trim (shell-command-to-string "sw_vers"))
-                 "\n## Editor\n"
-                 (emacs-version)))
-        ((or (eq system-type 'gnu/linux)
-             (eq system-type 'gnu/kfreebsd))
-         (concat text
-                 "\n# System info\n"
-                 "\n## OS details\n"
-                 (string-trim (shell-command-to-string "uname -a"))
-                 "\n## Editor\n"
-                 (emacs-version)))
-        ((eq system-type 'windows-nt)
-         (concat text
-                 "\n# System info\n"
-                 "\n## OS details\n"
-                 (string-trim (shell-command-to-string "ver"))
-                 "\n## Editor\n"
-                 (emacs-version)))
-        (t
-         (concat text
-                 "\n# System info\n"
-                 "\n## OS details\n"
-                 (format "%s" system-type)
-                 "\n## Editor\n"
-                 (emacs-version)))))
+  (cl-labels ((chatgpt-shell--get-system-info-command
+               ()
+               (cond ((eq system-type 'darwin) "sw_vers")
+                     ((or (eq system-type 'gnu/linux)
+                          (eq system-type 'gnu/kfreebsd)) "uname -a")
+                     ((eq system-type 'windows-nt) "ver")
+                     (t (format "%s" system-type)))))
+    (let ((system-info (string-trim
+                        (shell-command-to-string
+                         (chatgpt-shell--get-system-info-command)))))
+      (concat text
+              "\n# System info\n"
+              "\n## OS details\n"
+              system-info
+              "\n## Editor\n"
+              (emacs-version)))))
 
 (defcustom chatgpt-shell-system-prompts
   `(("tl;dr" . "Be as succint but informative as possible and respond in tl;dr form to my queries")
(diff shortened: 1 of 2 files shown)


xenodium (commented):
Thank you for the changes! I'm sorry it took me so long to review.

Generally looks good to me, though mostly wondering the goal of removing leading non-alpha-numeric characters from regions? 

Won't this yield some unexpected behaviour if we were to select something like the following?

```
#!/bin/bash
echo "hello"
```

How about we exclude the non-alphanumeric portion of the PR and keep the rest?

tninja:
Yes. Thanks for the feedback. Initially i just think the case for text instead of programming code. It is a good point with your example.

tninja:
Feedback addressed.

chatgpt-shell.el
```diff
@@ -1418,15 +1405,22 @@ enables additional key bindings.
...
 (defun chatgpt-shell-send-region-with-header (header)
   "Send text with HEADER from region using ChatGPT."
   (unless (region-active-p)
     (user-error "No region active"))
-  (chatgpt-shell-send-to-buffer
-   (concat header
-           "\n\n"
-           (buffer-substring (region-beginning) (region-end)))
-   nil))
+  (let* ((region-text (buffer-substring (region-beginning) (region-end)))
```
xenodium, line 1421: Let's keep the previous version. We aren't getting additional benefits from a local var, but also lose the explicitness of what previous code did.

chatgpt-shell.el
```diff
@@ -1418,15 +1405,22 @@ enables additional key bindings.
           (chatgpt-shell--primary-buffer)
         (chatgpt-shell--put-source-block-overlays)))))
 
+(defun chatgpt-shell-ask-chatgpt-region ()
```
xenodium, line 1408: Let's remove this function. Looks like it's missing some parts and we already have functions that do a similar thing:

`chatgpt-shell-send-region`
`chatgpt-shell-send-and-review-region`

chatgpt-shell.el
```diff
@@ -1476,11 +1470,12 @@ With prefix REVIEW prompt before sending to ChatGPT."
   (interactive "P")
   (unless (region-active-p)
     (user-error "No region active"))
-  (let ((chatgpt-shell-prompt-query-response-style 'shell))
+  (let* ((chatgpt-shell-prompt-query-response-style 'shell)
```
xenodium, line 1473: We can use `let` instead of `let*` in this case.

test_chatgpt-shell.el
```diff
@@ -0,0 +1,46 @@
...
+        (chatgpt-shell-model-versions '("gpt-3" "gpt-3.5" "gpt-4")))
+    ;; Test for nil, expecting nil
+    (should (equal (chatgpt-shell-model-version) nil))))
+
+;;; chatgpt-shell--append-system-info
+
+(ert-deftest test-chatgpt-shell--append-system-info-smoke-test ()
+  (let ((output (chatgpt-shell--append-system-info "abc")))
+    (should-not (null output))
+    ))
```
xenodium, line 26: nit: move parens to previous line

test_chatgpt-shell.el
```diff
@@ -0,0 +1,46 @@
...
+
+(defun mock-shell-command-to-string (command)
+  "A mock version of `shell-command-to-string` for testing purposes."
+  (concat "mocked system info"))
```
xenodium, line 30: Doesn't look like we need concat here? Maybe just: "mocked system info"

test_chatgpt-shell.el
```diff
@@ -0,0 +1,46 @@
...
+
+(defun mock-emacs-version ()
+  "A mock version of `emacs-version` for testing purposes."
+  "mocked emacs version")
+
+(ert-deftest test-chatgpt-shell--append-system-info-with-mock ()
+  (cl-letf (((symbol-function 'shell-command-to-string) #'mock-shell-command-to-string)
+            ((symbol-function 'emacs-version) #'mock-emacs-version)
+            )
```
xenodium, line 39: nit: parens in previous line

test_chatgpt-shell.el
```diff
@@ -0,0 +1,46 @@
...
+    (should (let ((system-type 'darwin))
+              (equal (chatgpt-shell--append-system-info "foo")
+                     "foo\n# System info\n\n## OS details\nmocked system info\n## Editor\nmocked emacs version")))
+    (should (let ((system-type 'gnu/linux))
+              (equal (chatgpt-shell--append-system-info "foo")
+                     "foo\n# System info\n\n## OS details\nmocked system info\n## Editor\nmocked emacs version")))
+            ))
```
xenodium, line 46: nit: parens in previous line

xenodium (commented):
Thank you! Looks like we have a few minor things left. Added comments.

chatgpt-shell.el
tninja: good point. will address

test_chatgpt-shell.el
tninja: will address

test_chatgpt-shell.el
tninja: yes correct. will remove concat

chatgpt-shell.el
tninja: hmm i am not quite agree with this one, cause nested expression might make debug / understanding difficult. breaking them into small piece might help understanding? my experience in other programming language let me made that change

tninja: also, i am still on the road learning emacs lisp. there is a built in (interactive "p") for region actions. wonder if it is good to move to use that way for existing region related functions in future, for simplicity? i could be wrong.

chatgpt-shell.el
tninja: oh sorry, that one need let* cause two local variable used there.

test_chatgpt-shell.el
tninja: addressed

test_chatgpt-shell.el
tninja: addressed

chatgpt-shell.el
xenodium: > nested expression might make debug / understanding / maintainance difficult.

The code we're considering here is trivial. Let's keep the previous please.

chatgpt-shell.el
xenodium: > oh sorry, that one need let* cause two local variable used there.

I'm only seeing `chatgpt-shell-prompt-query-response-style` and `region-text` declared and they don't seem to depend on each other. Why use `let*` instead of `let`?

xenodium (commented):
Thanks for the comments!

chatgpt-shell.el
tninja: i see. i thought let only allow one assignment instead multiple. Good to know this. i'll use let.

chatgpt-shell.el
tninja: OK, i can make it back

xenodium:
Hiya. I got notification to review changes. Seems like `chatgpt-shell-ask-chatgpt-region` is still there. Does the PR have the latest?

tninja:
Got it. just removed that function

xenodium (approved):
Thank you!

PR #198, opened 2024-03-03
Add default directory for restoring transcripts

djr7C4:
The function `chatgpt-shell-restore-session-from-transcript` does not current specify the directory when reading the filename of the transcript to restore. This can be a bit inconvenient for users as they will usually have to navigate to `shell-maker-transcript-default-path` manually each time. This PR fixes that issue.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -2053,7 +2053,7 @@ Very much EXPERIMENTAL."
   (interactive)
   (unless (eq major-mode 'chatgpt-shell-mode)
     (user-error "Not in a shell"))
-  (let* ((path (read-file-name "Restore from: " nil nil t))
+  (let* ((path (read-file-name "Restore from: " (file-name-as-directory shell-maker-transcript-default-path) nil t))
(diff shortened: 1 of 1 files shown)


chatgpt-shell.el
```diff
@@ -2053,7 +2053,7 @@ Very much EXPERIMENTAL."
   (interactive)
   (unless (eq major-mode 'chatgpt-shell-mode)
     (user-error "Not in a shell"))
-  (let* ((path (read-file-name "Restore from: " nil nil t))
+  (let* ((path (read-file-name "Restore from: " (file-name-as-directory shell-maker-transcript-default-path) nil t))
```
xenodium, line 2056: file-name-as-directory will cause an error out of the box as shell-maker-transcript-default-path is nil. We can likely just use shell-maker-transcript-default-path on its own here.

xenodium (commented):
Thanks for the PR. Left a comment to fix.

djr7C4:
Without `file-name-as-directory`, the path might not end in a slash which causes completion complete as a filename rather than complete the files located inside the directory. I've added an `and` which should fix the case when the path is nil.

PR #199, opened 2024-03-06
alias chatgpt-shell-historty-path to the new shell-maker variable

Vonfry:
commit 03afa73 updated `shell-maker-history-path` to `shell-maker-root-path`


Vonfry:
Or should we use `(shell-maker-history-file-path chatgpt-shell--config)` instead of a such variable?

xenodium (approved):
Sorry for the break and thank you for the fix!

xenodium:
On closer look, it should likely now be called `chatgpt-shell-root-path`. Modified in the latest.

Issue #208, opened 2024-06-13
option to re-use window for showing results of chatgpt-shell-prompt


solodov:
When  `chatgpt-shell-prompt-query-response-style` is set to `other-buffer` every prompt result is shown in a new window with a transient buffer in it. An option to reuse an existing window for the buffer would be great. This would allow to asynchronously send prompts without switching away from the main task and also not have emacs frame full of response windows.

I tried using `display-buffer-alist` for this:

```elisp
(add-to-list 'display-buffer-alist
             `(,(rx bol (| "*chatgpt* " (+ nonl) "> " (+ nonl))) .
               (display-buffer-reuse-window display-buffer-use-least-recent-window)))
```
This only works if prompt is identical, ie. results in the same buffer name as another transient response buffer already shown. A new prompt still produces new window.

xenodium:
Hi @solodov, thanks for the PR!

Trying things out... If I'm understanding correctly, you're using this to "send to shell" (like `*chatgpt* 4o/General`) and also "send to compose" (like `*chatgpt* 4o/General> compose`) buffers?

Would you mind capturing a screen grab of your preferred flow? Wanna make sure I'm seeing the behabiour you expect with the patch, but I also want avoid regressing other flows.

On a related note, I realized (retrospectively) that my existing code is being somewhat naughty by modifying `display-buffer-alist` on behalf of the user. I'd like to revisit that in the future and redesign a bit... While not directly applicable now, just a heads-up that things around buffer-handling may change in the future.

Btw, I'm thinking we can likely do without adding `chatgpt-shell-prompt-reuse-transient-buffer-window` and introduce a `other-buffer-reuse-window` style (or some other name, suggestions?) to `chatgpt-shell-prompt-query-response-style`. This way, we can prevent user misconfiguration if they forget to read the docstring stating `This variable only has effect when chatgpt-shell-prompt-query-response-style is set to other-window`.

If you want to skip the`other-buffer-reuse-window` change, I'm happy to follow up.

solodov:
Here's what this looks like. I'm calling `chatgpt-shell-prompt`. First half is with this setting turned off (current behaviour), second half with it turned on:

https://github.com/xenodium/chatgpt-shell/assets/187738/57241da5-5da7-4dcc-aeae-1de39dd9499b

[...]

PR #209, opened 2024-06-13
Introduce an option to re-use window showing transient buffer.

solodov:
Fixes xenodium/chatgpt-shell#208


steveyegge:
🎉🎉🎉 Woohoo!

xenodium:
As of https://github.com/xenodium/chatgpt-shell/commit/35437f283daa92946d549060b03ea4ac72d645cf, we should be able to reuse windows more idiomatically via `display-buffer-alist`. For example:

```emacs-lisp
(add-to-list 'display-buffer-alist
             (cons '(major-mode . chatgpt-shell-prompt-compose-mode)
                   '((display-buffer-reuse-window
                      display-buffer-in-direction
                      display-buffer-select)
                     (reusable-frames . visible)
                     (direction . left)
                     (window-width . 0.35))))
```

PR #213, opened 2024-06-22
Add insstructions for launching concurrent additional shells.

gavinhughes:
Add instructions to README for launching concurrent shells with =C-u M-x chatgpt-shell=.


xenodium (approved):
Thank you!

PR #232, opened 2024-10-03
new `chatgpt-shell-prompt-query-response-style` variant: 'replace

vzaliva:
A new `chatgpt-shell-prompt-query-response-style` variant: 'replace replaces currenty selected region with LLM output.

Example usage:

```
(defun chatgpt-shell-proofread-region-inline ()
  "Proofread English from region using ChatGPT."
  (interactive)
  (let ((chatgpt-shell-prompt-query-response-style 'replace))
    (chatgpt-shell-send-region-with-header
     (concat
      (cdr (assoc 'paper my-prompts))
      "\n Please respond with the corrected version only without any chatter, comments, or explanations.")
     )))
```

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -127,11 +127,13 @@
   "Determines the prompt style when invoking from other buffers.
 
 `'inline' inserts responses into current buffer.
+`'replace' inserts responses into current buffer replacing selected region.
 `'other-buffer' inserts responses into a transient buffer.
 `'shell' inserts responses and focuses the shell
 
 Note: in all cases responses are written to the shell to keep context."
   :type '(choice (const :tag "Inline" inline)
+                 (const :tag "Replace" replace)
                  (const :tag "Other Buffer" other-buffer)
                  (const :tag "Shell" shell))
   :group 'chatgpt)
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for this! I've been going back and forth about whether or not to extend the response styles. For some time, I've actually contemplated going the other way and removing global response style altogether and use function parameters instead. At the same time, this function was getting a little hairy, so I'm thinking of just modifying the behaviour to override region by default. This feels a little more natural anyway.

Does this work for your use-case?

```emacs-lisp
(defun chatgpt-shell-send-to-buffer (text &optional review handler on-finished)
  "Send TEXT to *chatgpt* buffer.
Set REVIEW to make changes before submitting to ChatGPT.

If HANDLER function is set, ignore `chatgpt-shell-prompt-query-response-style'

ON-FINISHED is invoked when the entire interaction is finished."
  (if (eq chatgpt-shell-prompt-query-response-style 'other-buffer)
      (let ((buffer (chatgpt-shell-prompt-compose-show-buffer text)))
        (unless review
          (with-current-buffer buffer
            (chatgpt-shell-prompt-compose-send-buffer))))
    (let* ((buffer (cond (handler
                          nil)
                         ((eq chatgpt-shell-prompt-query-response-style 'inline)
                          (current-buffer))
                         (t
                          nil)))
           (point (point))
           (marker (copy-marker (point)))
           (orig-region-active (region-active-p))
           (region-beginning (region-beginning))
           (region-end (region-end))
           (no-focus (or (eq chatgpt-shell-prompt-query-response-style 'inline)
                         handler)))
      (when (region-active-p)
        (setq marker (copy-marker (max (region-beginning)
                                       (region-end)))))
      (if (chatgpt-shell--primary-buffer)
          (with-current-buffer (chatgpt-shell--primary-buffer)
            (chatgpt-shell-start no-focus))
        (chatgpt-shell-start no-focus t))
      (cl-flet ((send ()
                  (when shell-maker--busy
                    (shell-maker-interrupt nil))
                  (goto-char (point-max))
                  (if review
                      (save-excursion
                        (insert text))
                    (insert text)
                    (shell-maker--send-input
                     (if (eq chatgpt-shell-prompt-query-response-style 'inline)
                         (lambda (_command output error finished)
                           (setq output (or output ""))
                           (when (buffer-live-p buffer)
                             (with-current-buffer buffer
                               (if error
                                   (unless (string-empty-p (string-trim output))
                                     (message "%s" output))
                                 (let ((inhibit-read-only t))
                                   (save-excursion
                                     (when orig-region-active
                                       (delete-region region-beginning region-end)
                                       (setq orig-region-active nil))
                                     (goto-char marker)
                                     (insert output)
                                     (set-marker marker (+ (length output)
                                                           (marker-position marker)))))))
                             (when (and finished on-finished)
                               (funcall on-finished))))
                       (or handler (lambda (_command _output _error _finished))))
                     t))))
        (if (or (eq chatgpt-shell-prompt-query-response-style 'inline)
                handler)
            (with-current-buffer (chatgpt-shell--primary-buffer)
              (goto-char (point-max))
              (send))
          (with-selected-window (get-buffer-window (chatgpt-shell--primary-buffer))
            (send)))))))
```

vzaliva:
I am fine with changing the default behaviour. I away from my computer and could not test it right now.

When working on my patch, I noticed that the second check `(eq ChatGPT 'inline)` did not work because the value of `ChatGPT` was changed mid-function. You may need to make sure this is not happening here.

xenodium:
Pushed https://github.com/xenodium/chatgpt-shell/commit/b76dcd4c15e98e3da98f9d5b1f0a0666a6165d5c

See if that works well for you. The proofread function should now replace (without explanation).

> You may need to make sure this is not happening here.

Hopefully that's resolved too.

vzaliva:
It seems to be working, but I will do more tests. Thanks!

xenodium:
I've added a new helper in https://github.com/xenodium/chatgpt-shell/commit/c9f1958e0a284432cb224ed9c5513eaf8e0c9d93, so writing one-off commands is easier but also does not pollute existing shell history with one-off query. I've migrated the proofreader, but the code more or less now looks like:

```emacs-lisp
(chatgpt-shell-request-and-insert-response
 :system-prompt "Please help me proofread the following English text and only reply with fixed text:"
 :query (if (region-active-p)
            (buffer-substring (region-beginning) (region-end))
          (error "No active region")))
```

Gonna go ahead and close the PR, but please reopen if you run into issues. Hope that's ok.

xenodium:
Btw, I've just added a generic modify command that seems to work for proofreading also. It's more visual of the changes it made.

proofread

vzaliva:
That looks just like what I need! Thanks! If only I can re-use some pre-defined prompt for that. My proofreading prompt is much longer than just "proofread".

xenodium:
> If only I can re-use some pre-defined prompt for that. My proofreading prompt is much longer than just "proofread".

Ok. Added in https://github.com/xenodium/chatgpt-shell/commit/1ed9e4a65a99964b1aef791876fc111734dc336f use the existing `chatgpt-shell-proofread-region` and customize with `chatgpt-shell-prompt-header-proofread-region`.

vzaliva:
Fantastic! Thanks. I will try it as soon as the new version hits MELPA. I am having a very busy week - no time to test from Git Hub. :(

vzaliva:
thanks!

PR #250, opened 2024-11-22
Custom ollama url base and source code block issue

ccqpein:
+ add ollama url base 
+ fix one source code block issue.

Sometimes the source code in "```" cannot match perfectly like: 

```
1. this is some thing
   and this is the line start from space
     ```elisp
      upper one is good 
      next "```" cannot match as the end of this source code block, because it start with spaces
      ```
```

So the highlight and the code copy/paste feature is confusing. 

(I am not sure I changed all "```" match in this PR, but I test with my case and it looks good. Feel free to change this PR)


xenodium (approved):
Thank you!

PR #258, opened 2024-11-26
Fix context length for OpenAI models

djr7C4:
The context length for OpenAI models is 128K according to OpenAI's documentation. However, the current code uses 12800 (12.8K) instead.


xenodium:
Oops. Thank you.

PR #259, opened 2024-11-26
Update context lengths for Google and Ollama models.

djr7C4:
These have been verified as per ollama show <model> and from the Google Gemini docs.


xenodium:
Thanks you!

PR #262, opened 2024-11-28
Initial implementation of chatgpt-shell-ollama-load-models

djr7C4:
This is a basic implementation of chatgpt-shell-ollama-load-models. I've tested it lightly and it seems to work at least for my ollama setup.

### DIFF
--- a/chatgpt-shell-ollama.el
+++ b/chatgpt-shell-ollama.el
@@ -109,6 +109,8 @@ VALIDATE-COMMAND handler."
                          :output)))
          (token-width (chatgpt-shell-ollama--parse-token-width
                        (map-elt (map-elt data 'details) 'quantization_level)))
+         ;; The context length key depends on the name of the model. For qwen2,
+         ;; it's at: model_info -> qwen2.context_length.
          (context-window (cdr (cl-find-if (lambda (cell)
                                             (string-suffix-p "context_length" (symbol-name (car cell))))
                                           (map-elt data 'model_info)))))
--- a/chatgpt-shell-ollama.el
+++ b/chatgpt-shell-ollama.el
@@ -120,9 +120,10 @@ VALIDATE-COMMAND handler."
 (cl-defun chatgpt-shell-ollama-load-models (&key override)
   "Query ollama for the locally installed models and add them to
 `chatgpt-shell-models' unless a model with the same name is
-already present. If OVERRIDE is non-nil (interactively with a
-prefix argument), remove all ollama models from
-`chatgpt-shell-models' before inserting locally installed models."
+already present. By default, replace the ollama models in
+`chatgpt-shell-models' locally installed ollama models. When
+OVERRIDE is non-nil (interactively with a prefix argument),
+replace all models with locally installed ollama models."
   (interactive (list :override current-prefix-arg))
   (let* ((ollama-predicate (lambda (model)
                              (string= (map-elt model :provider) "Ollama")))
@@ -130,8 +131,8 @@ prefix argument), remove all ollama models from
          ;; be placed in the same part of the list.
          (ollama-index (or (cl-position-if ollama-predicate chatgpt-shell-models)
                            (length chatgpt-shell-models))))
-    (when override
-      (setq chatgpt-shell-models (cl-remove-if ollama-predicate chatgpt-shell-models)))
+    (setq chatgpt-shell-models (and (not override)
+                                    (cl-remove-if ollama-predicate chatgpt-shell-models)))
     (let* ((existing-ollama-versions (mapcar (lambda (model)
                                                (map-elt model :version))
                                              (cl-remove-if-not ollama-predicate
--- a/chatgpt-shell-ollama.el
+++ b/chatgpt-shell-ollama.el
@@ -144,7 +144,10 @@ replace all models with locally installed ollama models."
       (setq chatgpt-shell-models
             (append (seq-take chatgpt-shell-models ollama-index)
                     new-ollama-models
-                    (seq-drop chatgpt-shell-models ollama-index))))))
+                    (seq-drop chatgpt-shell-models ollama-index)))
+      (message "Added %d ollama model(s); kept %d existing ollama model(s)"
+               (length new-ollama-models)
+               (length existing-ollama-versions)))))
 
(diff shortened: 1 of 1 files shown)


chatgpt-shell-ollama.el
```diff
@@ -84,6 +86,65 @@ VALIDATE-COMMAND handler."
...
+            (string-remove-suffix ":latest" (map-elt model 'name)))
+          (map-elt (shell-maker--json-parse-string
+                    (map-elt (shell-maker-make-http-request
+                              :async nil
+                              :url (concat chatgpt-shell-ollama-api-url-base "/api/tags"))
+                             :output))
+                   'models)))
+
+(defun chatgpt-shell-ollama--parse-token-width (quantization)
+  (string-match "^[FQ]\\([1-9][0-9]*\\)" quantization)
```
xenodium, line 100: We should prolly return nil if failed, so it can be handled later on.

```
(when-let ((matched (string-match "^[FQ]\\([1-9][0-9]*\\)" quantization)))
    (string-to-number (match-string 1 quantization)))
```

chatgpt-shell-ollama.el
```diff
@@ -84,6 +86,65 @@ VALIDATE-COMMAND handler."
...
+  (string-to-number (match-string 1 quantization)))
+
+(defun chatgpt-shell-ollama--make-model (version)
+  (let* ((data (shell-maker--json-parse-string
+                (map-elt (shell-maker-make-http-request
+                          :async nil
+                          :url (concat chatgpt-shell-ollama-api-url-base "/api/show")
+                          :data `((model . ,version)))
+                         :output)))
+         (token-width (chatgpt-shell-ollama--parse-token-width
```
xenodium, line 110: optionally: flatten with alist.

```
         (token-width (let-alist data
                        (chatgpt-shell-ollama--parse-token-width
                         .details.quantization_level)))
```

chatgpt-shell-ollama.el
```diff
@@ -84,6 +86,65 @@ VALIDATE-COMMAND handler."
...
+                              :async nil
+                              :url (concat chatgpt-shell-ollama-api-url-base "/api/tags"))
+                             :output))
+                   'models)))
+
+(defun chatgpt-shell-ollama--parse-token-width (quantization)
+  (string-match "^[FQ]\\([1-9][0-9]*\\)" quantization)
+  (string-to-number (match-string 1 quantization)))
+
+(defun chatgpt-shell-ollama--make-model (version)
```
xenodium, line 103: We should consider naming `chatgpt-shell-ollama--fetch-model`

chatgpt-shell-ollama.el
```diff
@@ -84,6 +86,65 @@ VALIDATE-COMMAND handler."
...
+  (let* ((data (shell-maker--json-parse-string
+                (map-elt (shell-maker-make-http-request
+                          :async nil
+                          :url (concat chatgpt-shell-ollama-api-url-base "/api/show")
+                          :data `((model . ,version)))
+                         :output)))
+         (token-width (chatgpt-shell-ollama--parse-token-width
+                       (map-elt (map-elt data 'details) 'quantization_level)))
+         (context-window (cdr (cl-find-if (lambda (cell)
```
xenodium, line 112: We could use a comment here. Maybe something like:

```
         ;; Context length key can vary. For qwen2, it's at:
         ;; model_info -> qwen2.context_length
```

chatgpt-shell-ollama.el
```diff
@@ -84,6 +86,65 @@ VALIDATE-COMMAND handler."
...
+                                             (cl-remove-if-not ollama-predicate
+                                                               chatgpt-shell-models)))
+           (new-ollama-versions (cl-remove-if (lambda (version)
+                                                (member version existing-ollama-versions))
+                                              (chatgpt-shell-ollama--fetch-model-versions)))
+           (new-ollama-models (mapcar #'chatgpt-shell-ollama--make-model new-ollama-versions)))
+      (setq chatgpt-shell-models
+            (append (seq-take chatgpt-shell-models ollama-index)
+                    new-ollama-models
+                    (seq-drop chatgpt-shell-models ollama-index))))))
```
xenodium, line 146: We likely want some sort of user feedback:

```
      (setq chatgpt-shell-models
            (append (seq-take chatgpt-shell-models ollama-index)
                    new-ollama-models
                    (seq-drop chatgpt-shell-models ollama-index)))
      (message "Ollama: added %d model(s), cached %d model(s)"
               (length new-ollama-models)
               (length existing-ollama-versions))
```

chatgpt-shell-ollama.el
```diff
@@ -84,6 +86,65 @@ VALIDATE-COMMAND handler."
...
+prefix argument), remove all ollama models from
+`chatgpt-shell-models' before inserting locally installed models."
+  (interactive (list :override current-prefix-arg))
+  (let* ((ollama-predicate (lambda (model)
+                             (string= (map-elt model :provider) "Ollama")))
+         ;; Find the index of the first ollama model so that the new ones will
+         ;; be placed in the same part of the list.
+         (ollama-index (or (cl-position-if ollama-predicate chatgpt-shell-models)
+                           (length chatgpt-shell-models))))
+    (when override
```
xenodium, line 133: Oooh, your override implementation is much nicer than I had in mind. I was thinking of just overriding the entire list (wiping non-Ollama models) without considering merging. I like your merge strategy more.

How about we combine both approaches and:

:override t = override entire list (wipe non-Ollama models)
:override nil = merge Ollama models.

ps. Not sure if you've seen, but we have `M-x chatgpt-shell-reload-default-models`, which is handy for re-instating default models.

xenodium:
> Currently, when override is t only the existing ollama models are removed
before local models found in the system are added. I would rather not
remove this behavior as I am using it.

Oh yes. I'm suggesting we make this the default command behaviour (:override nil).

:override t would go futher and replace the entire content of the list with fetched models (no merging at all, simply wipe and replace). Hope that makes more sense.

djr7C4:
Oh I see. That sounds reasonable given your point that we already have a
command that restores the default models.

djr7C4:
@xenodium I've added the changes you proposed above.

xenodium (approved):
Looks great! Thank you for the changes.

PR #271, opened 2024-12-08
Roll own csv parsing, remove `pcsv` dependency

arminfriedl:
Parse the CSV retrieved from Awesome Prompts [1] on our own. This
removes the dependency on the pcsv package which is not standard.

While there is certainly an argument to be made against NIH, CSV parsing
should also be do-able with reasonable added complexity, at least if the
format is somewhat well-known. A dependency on a third party package
just for this seems unnecessary.

The pcsv parser was added in response to issue #126. I am using this parser
suggested here in claude-shell [2] since quite a while and it worked
well enough - and as far as I can tell also correct.

[1] https://raw.githubusercontent.com/f/awesome-chatgpt-prompts/main/prompts.csv
[2] https://github.com/arminfriedl/claude-shell

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -422,7 +422,28 @@ Or nil if none."
   (chatgpt-shell-interrupt nil)
   (chatgpt-shell--save-variables))
 
-(declare-function pcsv-parse-file "pcsv" (file &optional coding-system))
+(defun chatgpt-shell--load-awesome-prompts-parse-alist ()
+  "Helper function for `claude-shell-load-awesome-prompts'.
+
+Download awesome-prompts and parse into a list of label and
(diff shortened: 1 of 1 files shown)


xenodium:
Sounds good, though I may reach to you for maintenance if follow-ups are needed, if that's alright?

xenodium (approved):

xenodium:
Thank you!

> Sounds good, though I may reach out to you for maintenance if follow-ups are needed, if that's alright?

Merged with this caveat ☝️

arminfriedl:
Thank you!

Yes of course, that's alright. I figure this is a small enough snippet that I'm happy to maintain this little part for the foreseeable future :hammer: :wink:

PR #286, opened 2024-12-31
Put the whole chatgpt shell buffer name inside the earmuffs

djr7C4:
This is for consistency with other shell-style buffers in Emacs


xenodium (approved):
😅 okay okay

Thanks for the change.

djr7C4:
Yeah, I'm picky (at least apart from checkdoc stuff I guess 🤣).

PR #290, opened 2025-01-01
Make chatgpt-shell respect `chatgpt-shell-root-path`

arminfriedl:
`chatgpt-shell-root-path` is already defined as an alias for `shell-maker-root-path`. I'm per-se fine with that, although I could also see that some might prefer it even independently configurable. In any case, this is currently unused.

Make chatgpt respect the `chatgpt-shell-root-path` when creating internal files like `.chatgpt-shell.el`.

Use case: I want to keep a tidy `emacs.d` with my configuration and auto-generated files clearly separated.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -2999,7 +2999,7 @@ compiling source blocks."
   "Save variables across Emacs sessions."
   (setq-default chatgpt-shell-system-prompt
                 chatgpt-shell-system-prompt)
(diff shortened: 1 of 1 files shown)


xenodium (approved):

xenodium:
Thank you. I may have missed this during a refactor. Either way, thanks for looking into it and fixing it.

arminfriedl:
Thanks to you and happy new year! :sparkle:

PR #303, opened 2025-02-06
Add grounding search feature to gemini

ccqpein:
~`(setf chatgpt-shell-google-enable-grounding-search t)` can turn on the google search for gemini models~

`gemini-2.0-flash` can use grounding search

### DIFF
--- a/chatgpt-shell-google.el
+++ b/chatgpt-shell-google.el
@@ -48,11 +48,6 @@ If you use Gemini through a proxy service, change the URL base."
   :safe #'stringp
   :group 'chatgpt-shell)
 
-(defcustom chatgpt-shell-google-enable-grounding-search nil
-  "Enable the grounding google search tool for gemini models."
-  :type 'boolean
-  :group 'chatgpt-shell)
-
 ;; https://ai.google.dev/gemini-api/docs/tokens
 ;; A token is equivalent to _about_ 4 characters.
 (cl-defun chatgpt-shell-google-make-model (&key version short-version path token-width context-window grounding-search)


chatgpt-shell-google.el
```diff
@@ -189,8 +194,10 @@ or
                    (chatgpt-shell-google--gemini-user-model-messages
                     (append context
                             (when prompt
-                              (list (cons prompt nil)))))))
-     (generation_config . ((temperature . ,(or temperature 1))
+                              (list (cons prompt nil))))))))
+   (when chatgpt-shell-google-enable-grounding-search
+     '((tools . ((google_search . ())))))
```
ccqpein, line 199: Just add `{"tools": {"google_search": {}}}` to request body can enable search:

ref: https://github.com/google-gemini/cookbook/blob/main/quickstarts/Search_Grounding.ipynb

xenodium: Grounding works great! I'm thinking we should make this feature part of each model instantiation to make sure we don't mix and match unsupported features across Google models. This also paves the way to export the feature more generically.

We can initially turn it on by default for the Google models that support grounding with something like this:

https://gist.github.com/xenodium/e52fb4dba03d47a83d35379e21623238

I reckon we can do it like this as a stepping stone to be merged in this PR.

We can follow up with another PR later on (no rush)  with a general setting to turn grounding on/off, similar to how we do system prompts or temperature.

Another interim option is to instantiate the Google groundable models twice (one with :grounding-search t and one :grounding-search nil). They also need different names of course.

Does that sound ok?

Feel free to use the gist for this PR if keen, otherwise I'm happy to follow-up.

ccqpein: I am ok with your change. I think it is ok that make this feature for special models, then we give the turn on/off function to current model.

xenodium: Nice. Do you want to apply it or should I?

ccqpein: @xenodium I just update with your diff. Please check it out.

xenodium (approved):

xenodium:
Thanks for the iteration! Merging

DinoChiesa:
[PR #304](https://github.com/xenodium/chatgpt-shell/pull/304/files#diff-ad40fa27bc79d40067d1dd26b6a1437325945c9a4befb807722a3876757b177fR171) includes a fn to toggle grounding for the current, in-use Gemini model. 

https://github.com/user-attachments/assets/b6ea5c37-61a3-4545-92b8-9bcdb6473907

djr7C4:
I use 2.5 pro and wouldn't like to see it removed. IMO purple shouldn't use the API if they are unaware of how billing works. Charges are the expectation.

PR #308, opened 2025-02-25
Add o3-mini models (which are now usable through OR)

djr7C4:
o1, o3-mini and o3-mini-high are now all useable through OpenRouter without a high-teir OpenAI API key (which was previously required). A high-teir API key still seems to be required to use them through OpenAI directly.

### DIFF
--- a/chatgpt-shell-openrouter.el
+++ b/chatgpt-shell-openrouter.el
@@ -77,6 +77,22 @@ VALIDATE-COMMAND and OTHER-PARAMS for `chatgpt-shell-openai-make-model'."
          ;; See https://openrouter.ai/qwen/qwq-32b-preview
          :other-params '((provider (quantizations . ["bf16"]))))
         (chatgpt-shell-openrouter-make-model
+         :version "openai/o3-mini-high"
+         :short-version "o3-mini-high"
+         :label "ChatGPT"
+         :token-width 3
+         ;; See https://openrouter.ai/openai/o3-mini-high
+         :context-window 200000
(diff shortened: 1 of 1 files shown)


djr7C4:
I should also mention that o3-mini takes a parameter that controls the amount of reasoning it performs when responding to the prompt. o3-mini uses the default value while o3-mini-high allows for more reasoning but will also result in higher costs.

https://openrouter.ai/openai/o3-mini-high

xenodium (approved):

xenodium:
Thanks for the PR to add the models! Merged.

xenodium:
> I should also mention that o3-mini takes a parameter that controls the amount of reasoning it performs when responding to the prompt.

Since the chatgptp-shell split, we could handle these in model-specific instantiations (ie. chatgpt-shell-openrouter-make-model), though we'd need a little piping work to expose as a user-setting we can easily swap at run-time (like we do for say system prompt). Anyway, not saying we need to do this now. Just that at some point it may be useful to have.

djr7C4:
Yes. We could also add a configuration variable that controls the default reasoning level (which can be overridden by an argument to the function.

djr7C4:
I looked into this because I wanted to use `o1` with a `reasoning_effort` set to `"high"`. Unfortunately, while it is documented by OpenAI, they don't actually accept `reasoning_effort` in actual API calls so far as I can tell. The code is here if you are interested: https://github.com/djr7C4/chatgpt-shell/tree/reasoning-parameter

PR #312, opened 2025-02-28
Use the latest versions for Anthropic models instead of dated versions

djr7C4:
Like OpenAI and Google, Anthropic provides models with the suffix "-latest" that point to the newest version of each of their models (e.g. Sonnet-3.7, Sonnet-3.5, etc.). Avoiding using versions tagged with a specific date reduces the need to update the models as new releases are made and is in line with what is already in place for Google and OpenAI models.


xenodium (approved):
Much better. Thank you.

PR #316, opened 2025-03-06
Add the ability to ignore some model versions when swapping

djr7C4:
This is useful for keeping redundant models (like openrouter "o3-mini" which is also available more cheaply through openai) in the models list but enabling users to hide them if they don't need them at the moment.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -521,9 +521,9 @@ Downloaded from https://github.com/f/awesome-chatgpt-prompts."
   (setq chatgpt-shell-models (chatgpt-shell--make-default-models))
   (message "Reloaded %d models" (length chatgpt-shell-models)))
 
-(defcustom chatgpt-shell-model-filter (lambda (_model) t)
-  "A predicate used to determine which models should be shown in
-`chatgpt-shell-swap-model'."
+(defcustom chatgpt-shell-model-filter nil
+  "A function that is applied `chatgpt-shell-models' to determine
+which models should be shown in `chatgpt-shell-swap-model'."
   :type 'function
   :group 'chatgpt-shell)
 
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for the PR.

I can foresee folks also wanting the opposite (allowlisting) or even specialised filter logic.

Mind making this a little more generic? That is, introduce a chatgpt-shell-model-filter variable that takes a filter function. This variable be used to filter with whatever logic users prefer, including the ignore proposal.

djr7C4:
How do you envision the allow logic working given that everything is
already allowed by default?

xenodium:
The filter logic to "keep". Everything else discarded.

djr7C4:
I've added the variable you suggested. I decided to add functions that return closures to use a predicates as it seems cleaner than having ignore/allow list variables that are only in effect when chatgpt-shell-model-filter is set to the right function.

xenodium (approved):
Thanks for this! I was thinking of applying the function directly. For example:

```diff
                                        (map-elt model :provider)
                                        (map-elt model :version)))
-                             chatgpt-shell-models))
+                             (if chatgpt-shell-model-filter
+                                 (apply chatgpt-shell-model-filter chatgpt-shell-models)
+                                 chatgpt-shell-models)))
             (selection (nth 1 (split-string (completing-read "Model version: "
                                                              models nil t)))))
```

The above may be a little more versatile (you can use something other than seq-filter), though I'm equally happy with the predicate proposal, which I can merge if you want to go through with it.

xenodium:
> I'm equally happy with the predicate proposal, which I can merge if you want to go through with it.

If we do keep the predicate, maybe we name the variable chatgpt-shell-model-filter-predicate? WDYT?

djr7C4:
Your approach is more versatile and I can imagine some use cases for it (though my current one does not require it). I'd say there's no reason not to apply the function directly.

djr7C4:
Changes have been pushed to call the filter function directly.

xenodium (approved):

xenodium:
Thanks for the changes. Merged.

xenodium:
ps. If keen, re-run the README.org code block to generate variables's documentation at https://github.com/xenodium/chatgpt-shell/blob/main/README.org?plain=1#L237. Otherwise I'll do it shortly.

djr7C4:
Done. This could also be done automatically on GitHub (or maybe via a git hook).

PR #317, opened 2025-03-08
added anthropic to open router

italoaa:
Simple add. Did it mostly because I wanted to use anthropic through open router and though it should be good to contribute to this nice project :)


xenodium:
Thank for the PR. Merged.

PR #319, opened 2025-03-10
feat: sane sort for swap-model with icomplete-vertical

DinoChiesa:
I use [icomplete-vertical](https://github.com/oantolin/icomplete-vertical).  It sorts candidates in the minibuffer first by length, then alphabetically. This makes for a nonsensical experience when swapping models in chatgpt-shell.  (For uses that employ a different completion framework or plugin, the sorting might be different)

With this change, sorting of model names during swap-model is sane.

Before:
emacs_bJDKpwXCO5

After:
emacs_G2x3aDf6k5

### DIFF
--- a/chatgpt-shell-google.el
+++ b/chatgpt-shell-google.el
@@ -1,6 +1,6 @@
 ;;; chatgpt-shell-google.el --- Google-specific logic  -*- lexical-binding: t -*-
 
-;; Copyright (C) 2023 Alvaro Ramirez
+;; Copyright (C) 2023-2025 Alvaro Ramirez
 
 ;; Author: Alvaro Ramirez https://xenodium.com
 ;; URL: https://github.com/xenodium/chatgpt-shell
@@ -27,8 +27,11 @@
 
 (eval-when-compile
   (require 'cl-lib))
+(require 'let-alist)
 (require 'shell-maker)
 (require 'map)
+(require 'rx)
+(require 'json)
 
 (defvar chatgpt-shell-proxy)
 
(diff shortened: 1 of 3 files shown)


djr7C4:
This will be an improvement for some users. However, for some completion frameworks it may not be depending on the user's preferences. 

With `vertico` for example, candidates are sorted using heuristics when a sort function is not provided in the competition table and more recently chosen candidates appear first. This pull request would override that behavior to always force the candidates to appear in the same order.

I suggest that this be made an optional feature that can be toggled using a variable.

xenodium:
Thanks for the proposal @DinoChiesa 

> I use [icomplete-vertical](https://github.com/oantolin/icomplete-vertical). It sorts candidates in the minibuffer first by length, then alphabetically.

This is interesting. I'm using ivy and order is shown in the same order as the `chatgpt-shell-models` variable. 

image

There's benefit in using the default order (as per `chatgpt-shell-models`). At present, entries are more or less ordered by provider/model (ie. Anthropic, DeepSeek, ...) and recency. Newer models are typically added at the top (of their provider category). This makes the newer models more discoverable. I've kinda lost track of what the latest models are for all the different providers, but thankfully with contributors adding new models, I get to discover new ones since they surface to top of their category.

> I suggest that this be made an optional feature that can be toggled using a variable.

@djr7C4 recently added a `chatgpt-shell-model-filter` variable. While it's meant for filtering prior to swapping, it's effectively a mechanism to apply any function to `chatgpt-shell-models` prior to swapping. I don't know much about  [icomplete-vertical](https://github.com/oantolin/icomplete-vertical), but I'm wondering if `chatgpt-shell-model-filter` would suffice for your use-case? Can you pre-sort using that?

djr7C4:
@xenodium 

The problem with your suggestion is that it will not result in consistent behavior across different completion frameworks. Some frameworks do not preserve the original order whereas others do. 

One option would be to make the sort function itself configurable via a variable. When the sort function is set in the completion table (display-sort-function in the code) it should be consistent across all frameworks that comply with the Emacs documentation. cycle-sort -function serves a slightly different function and likely won't be needed in most cases but for correctness and completeness it's worth setting as well.

DinoChiesa:
I don't know about the variations in completion frameworks; I figured this PR would foster a discussion. 

> I'm wondering if chatgpt-shell-model-filter would suffice for your use-case? Can you pre-sort using that?

Pre-sorting does not work. The _only_ way I found to get sorted items in `completing-read` was to provide a function for the COMPLETIONS argument. If callers do not provide that, then there is no sort function, and `completion-all-sorted-completions` applies its own sort. The default they chose here seems... surprising.
Screenshot 2025-03-10 9 14 51 AM

[THIS SO thread](https://emacs.stackexchange.com/a/8177/3856) suggested the approach I took. 
Obviously ivy and other frameworks are sorting differently.  Maybe icomplete-vertical is the outlier and isn't doing what these other frameworks do. I don't understand the difference; maybe one of you do. 

> One option would be to make the sort function itself configurable via a variable. 

Yes, maybe that would work.  Introduce `chatgpt-shell-model-sort` as a fn, that people can set. Would be useful for icomplete-vertical, but people could leave it unset to get whatever sorting they have in their chosen completions framework.

```
(defun chatgpt-shell--model-swap-sorter (candidates)
   "Sorts model CANDIDATES if the user has provided a
sort function."
   (if chatgpt-shell-model-sort
       (funcall chatgpt-shell-model-sort candidates)
     (identity candidates)))
```
And then, in user code, a person could use this:
```
(setq chatgpt-shell-model-sort (lambda (candidates)  (sort candidates :lessp #'string<)))
```

djr7C4:
> > One option would be to make the sort function itself configurable via a variable. 
> 
> Yes, maybe that would work.  Introduce `chatgpt-shell-model-sort` as a fn, that people can set. Would be useful for icomplete-vertical, but people could leave it unset to get whatever sorting they have in their chosen completions framework.
> 
> ```
> (defun chatgpt-shell--model-swap-sorter (candidates)
>    "Sorts model CANDIDATES if the user has provided a
> sort function."
>    (if chatgpt-shell-model-sort
>        (funcall chatgpt-shell-model-sort candidates)
>      (identity candidates)))
> ```
> And then, in user code, a person could use this:
> ```
> (setq chatgpt-shell-model-sort (lambda (candidates)  (sort candidates :lessp #'string<)))
> ```

I would instead set chatgpt-shell-model-sort as the display-sort-function in the completion table. Otherwise, there isn't a way to leave display-sort-function unset and let the completion framework choose its own sorting function. This would prevent vertico users from choosing to use vertico's heuristics instead of the default sort order.

djr7C4:
> The only way I found to get sorted items in completing-read was to provide a function for the COMPLETIONS argument

Right. If you don't provide this as metadata in the completion table than the default behavior depends on what framework you are using.

djr7C4:
BTW, I the `lexical-let` is unnecessary because `lexical-binding: t` is set locally already in the source file. This is an old macro that was used before `lexical-binding` was added to Emacs.

DinoChiesa:
> I would instead set chatgpt-shell-model-sort as the display-sort-function in the completion table. Otherwise, there isn't a way to leave display-sort-function unset and let the completion framework choose its own sorting function. This would prevent vertico users from choosing to use vertico's heuristics instead of the default sort order.

ok, so, 
- specify the COMPLETIONS function as I have done (but without the lexical-let)
- the lambda will return 
  ```
           (cons 'metadata
                 (list
                  (cons 'cycle-sort-function
                        #'chatgpt-shell-model-sort)
                  (cons 'display-sort-function
                        #'chatgpt-shell-model-sort)))
   ```
   ...for metadata. 
- just leave `chatgpt-shell-model-sort` as nil .  And if people want to override it, they can do so.
- no need for a `chatgpt-shell--model-swap-sorter` fn

is that about right?

xenodium:
Hiya folks! Sorry was on a plane. Catching up...

How about we go a little higher-level and rather than deal with sorting specifically, we provide a configurable `chatgpt-shell-completing-read-swap-function`? This gives the full flexibility to use whatever concrete completing function (ie. ivy-read, helm etc, but also completing-read with whatever configuration/params anyone may want.

djr7C4:
That would work as well.

DinoChiesa:
I'm not sure what additional problems that would solve. The current impl uses `competing-read`. I thought all the frameworks would hook into that, in some way.  With this PR I was trying to work within that landscape.

I've updated the PR so that `chatgpt-shell-model-sort` is a defcustom as summarized in [my prior comment](https://github.com/xenodium/chatgpt-shell/pull/319#issuecomment-2711810145). See what you think. I have no strong attachment to this; it's just a suggestion.

djr7C4:
Some completion frameworks have options that are available directly through their completion functions (e.g. `helm-comp-read` and `ivy-read`) support options that cannot be controlled via `completing-read` so this approach is a bit more flexible. `helm-comp-read` in particular has a huge number of keyword arguments that allow for fine-grained control over completion.

DinoChiesa:
Is the spirit of over-engineering creeping into this model swap challenge?  I suppose if the goal is to support many, many more models than are currently available, 1000's? then it will begin to make sense to want finer control over completion.  But even then, I am imagining that with most frameworks, typing just a few characters narrows the set of choices rapidly.  Google now lists ~28 models that support "generateContent".  When I load those in, the total number of models chatgpt-shell is handling is ~55.

I understand in general why ivy is valuable, and also... I don't see the need for that sort of capability in this small enclave of capability. But that's my opinion. Maybe there is some aspect of UX consistency I am neglecting; maybe people who use ivy find it strange and jarring to have to use completing-read.  That would be a good reason to generalize it I guess. 

As I said I don't have strong attachments to this PR. I can monkey patch the module to make it do what I want.

DinoChiesa:
ok I get it. 

I filed a different PR, see #321 .

xenodium:
> I filed a different PR, see https://github.com/xenodium/chatgpt-shell/pull/321 .

Thanks for iterating on it. #321 now merged.

PR #323, opened 2025-03-19
Add experimental version of gemini 2.0 pro and use latest versions

djr7C4:
Previously experimental models for gemini were only available with the date in the model version. However, it is now possible to always use the latest version.


xenodium (approved):
Thank you

PR #328, opened 2025-03-24
Change some openai models and add new models

ccqpein:
Just noticed there are several changes on openai API side:

+ `chatgpt-4o-latest` document link change to [https://platform.openai.com/docs/models/chatgpt-4o-latest](https://platform.openai.com/docs/models/chatgpt-4o-latest) 
+ [https://platform.openai.com/docs/models/gpt-4o](https://platform.openai.com/docs/models/gpt-4o) link to the other model gpt-4o

Several changes:

+ `gpt-4o` is much cheaper, so I move it to be the first option than `chatgpt-4o-latest`
+ also add the `gpt-4o-mini`
+ added two models those can search web `gpt-4o-search-preview` and `gpt-4o-mini-search-preview`

### DIFF
--- a/chatgpt-shell-openai.el
+++ b/chatgpt-shell-openai.el
@@ -71,6 +71,11 @@ HANDLER, FILTER and OTHER-PARAMS."
   "Build a list of all OpenAI LLM models available."
   ;; Context windows have been verified as of 11/26/2024.
   (list (chatgpt-shell-openai-make-model
+         :version "chatgpt-4o-latest"
+         :token-width 3
+         ;; https://platform.openai.com/docs/models/chatgpt-4o-latest
+         :context-window 128000)
+        (chatgpt-shell-openai-make-model
          :version "gpt-4o"
          :token-width 3
          ;; https://platform.openai.com/docs/models/gpt-4o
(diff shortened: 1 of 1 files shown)


djr7C4:
FWIW, I prefer the current order as chatgpt-4o-latest points to a newer version of gpt-4o (gpt-4o-2024-11-20 vs gpt-4o-2024-08-06). I find the difference in price negligible unless the models are being used with very large inputs. The order in `chatgpt-shell-swap-model` can also be changed using `chatgpt-shell-swap-model-filter`.

ccqpein:
@djr7C4 I am not sure the `chatgpt-4o-latest` point to `gpt-4o`. In document, it point to `chatgpt-4o`. I don't know if the `chatgpt-4o` is the `gpt-4o` (`gpt-4o-2024-11-20`).

I click the `gpt-4o-2024-11-20` in 4o document, the price is the same as `gpt-4o-2024-08-06`. Half cheaper than `chatgpt-4o-latest`.

djr7C4:
The documentation on the chatgpt-4o-latest page only states that "ChatGPT-4o points to the GPT-4o snapshot currently used in ChatGPT." However, based on the name, it wouldn't make sense if it pointed to a snapshot older than gpt-4o-2024-11-20.

gpt-4o does not point to gpt-4o-2024-11-20. It points to gpt-4o-2024-08-06. 

IMO, the cost doesn't make that much of a difference unless you are using the models with massive inputs. In my use case, it's a few dollars a month.

ccqpein:
> gpt-4o does not point to gpt-4o-2024-11-20. It points to gpt-4o-2024-08-06.

Yes, I manually click it and check if the price is same or not. 

I think the price is ok too. Let me make chatgpt latest as the first.

xenodium (approved):
Thank you for the PR!

PR #329, opened 2025-03-24
Add support for the reasoning_effort parameter

djr7C4:
If we add an interface for setting parameters such as reasoning_effort, this would allow models such as o3-mini-high to be removed as redundant.


djr7C4:
Note that o3-mini was added back in when main was merged into this branch.

chatgpt-shell-openai.el
```diff
@@ -79,6 +91,7 @@ HANDLER, FILTER and OTHER-PARAMS."
          :version "o3-mini"
          :token-width 3
          :context-window 200000
+         :reasoning-effort chatgpt-shell-openai-default-reasoning-effort
```
xenodium, line 94: I'm wondering if we should consider deferring access to the actual level value until the the payload is constructed? This would enable changing `chatgpt-shell-openai-default-reasoning-effort` at any time and be readily usable (without having to reload the models).

For some values, we've achieved this by setting the value to a function that returns the latest value of a variable with the same name.

chatgpt-shell-openai.el
```diff
@@ -34,7 +34,18 @@
 (declare-function chatgpt-shell--make-chatgpt-url "chatgpt-shell")
 (declare-function chatgpt-shell-validate-no-system-prompt "chatgpt-shell")
 
-(cl-defun chatgpt-shell-openai-make-model (&key version short-version token-width context-window validate-command (headers #'chatgpt-shell-openai--make-headers) (key chatgpt-shell-openai-key) (url-base 'chatgpt-shell-api-url-base) (path "/v1/chat/completions") (provider "OpenAI") (label "ChatGPT") (handler #'chatgpt-shell-openai--handle-chatgpt-command) (filter #'chatgpt-shell-openai--filter-output) other-params)
+;; See https://platform.openai.com/docs/guides/reasoning
+(defcustom chatgpt-shell-openai-default-reasoning-effort "medium"
```
xenodium, line 38: Should this be buffer local? It would enable setting level per shell/buffer.

chatgpt-shell-openai.el
```diff
@@ -91,18 +104,23 @@ HANDLER, FILTER and OTHER-PARAMS."
...
          :context-window 200000
+         :reasoning-effort chatgpt-shell-openai-default-reasoning-effort
          :validate-command #'chatgpt-shell-validate-no-system-prompt)
         (chatgpt-shell-openai-make-model
          :version "o1-preview"
          :token-width 3
          ;; https://platform.openai.com/docs/models/gpt-01
          :context-window 128000
+         ;; Reasoning effort is only supported for o1-pro, o1 and o3-mini.
+         ;; :reasoning-effort chatgpt-shell-openai-default-reasoning-effort
```
xenodium, line 115: Is this meant to be commented out? If so, perhaps consider `:reasoning-effort nil`

chatgpt-shell-openai.el
```diff
@@ -91,18 +104,23 @@ HANDLER, FILTER and OTHER-PARAMS."
...
          :validate-command #'chatgpt-shell-validate-no-system-prompt)
         (chatgpt-shell-openai-make-model
          :version "o1-mini"
          :token-width 3
          ;; https://platform.openai.com/docs/models/gpt-01-mini
          :context-window 128000
+         ;; Reasoning effort is only supported for o1 and o3-mini.
```
xenodium, line 122: Ditto

chatgpt-shell-openai.el
```diff
@@ -206,6 +224,10 @@ and OTHER-PARAMS (list)."
                             :context context))))
    (when temperature
      `((temperature . ,temperature)))
+   ;; Unfortunately, while it is documented by OpenAI, the reasoning_effort
```
xenodium, line 227: I didn't quite understand the comment. Does it mean the patch doesn't currently work as expected?

chatgpt-shell-openrouter.el
```diff
@@ -75,31 +77,36 @@ VALIDATE-COMMAND and OTHER-PARAMS for `chatgpt-shell-openai-make-model'."
...
-         :other-params '((provider (quantizations . ["bf16"]))))
-        (chatgpt-shell-openrouter-make-model
-         :version "openai/o3-mini-high"
-         :short-version "o3-mini-high"
-         :label "ChatGPT"
-         :token-width 3
-         ;; See https://openrouter.ai/openai/o3-mini-high
-         :context-window 200000
-         :validate-command #'chatgpt-shell-validate-no-system-prompt)
+         :other-params '((provider (quantizations . ["bf16"])
```
xenodium, line 80: Don't we need to delete altogether?

xenodium (commented):
Thanks a lot for the PR! Sorry for the delay. Left some comments.

chatgpt-shell-openai.el
djr7C4: That is one option. My thought was that in the future we might have a UI for changing parameters like this on a per model basis. In that case, it makes more sense for it to be a default as it is currently. However, if we decide not to do that then your way makes more sense.

chatgpt-shell-openai.el
djr7C4: It is meant to be commented out. :reasoning-effort is already nil by default when not provided. I wanted a comment here so someone wouldn't think it should be added in the future.

chatgpt-shell-openai.el
djr7C4: It should be working now. Previously, when I tested it the API didn't accept the parameter (that's why this branch h is a bit old and I didn't submit it before). However, OpenAI seems to have fixed that now.

chatgpt-shell-openrouter.el
djr7C4: This is for qwen and not o3-mini-high. It looks like it ended up in the diff due to the parens unless I'm missing something here.

chatgpt-shell-openai.el
djr7C4: I guess the two aren't mutually exclusive anyway. The UI could just have a list of buffer-local variables that can be modified and work that way so probably this change is the way to go anyway.

djr7C4:
The :reasoning-effort key for models is now a boolean. When non-nil, `chatgpt-shell-openai-reasoning-effort` is used as the value for the reasoning effort. The variable isn't buffer-local currently but if we add a UI for editing parameters later, we can just use `setq-local` in it as `chatgpt-shell-swap-model` does now.

chatgpt-shell-openrouter.el
xenodium: I mean the block that's commented out. Don't we need to get rid of it?

        ;; (chatgpt-shell-openrouter-make-model
        ;;  :version "openai/o3-mini-high"
        ;;  :short-version "o3-mini-high"
        ;;  :label "ChatGPT"
        ;;  :token-width 3
        ;;  ;; See https://openrouter.ai/openai/o3-mini-high
        ;;  :context-window 200000
        ;;  :validate-command #'chatgpt-shell-validate-no-system-prompt)

xenodium:
Thanks for the latest changes. Replied with some minor things. I can get to them later today before merging, unless you want to give them a go.

chatgpt-shell-openrouter.el
djr7C4: Feel free if you prefer. I sometimes leave such things commented out.

chatgpt-shell-openai.el
djr7C4: I prefer the commented code but this is your project so I won't argue 🙂

xenodium (approved):

xenodium:
Thanks for the PR! Merged.

PR #334, opened 2025-03-24
fix: use system_instruction conditionally

DinoChiesa:
This fix embeds the system_instruction portion of the POST body , conditionally. 

With this fix, using most Google-hosted models, the POST body looks like: 

```
{
  "system_instruction": {
    "parts": {
      "text": "You use markdown liberally to structure responses. Always show code snippets in markdown blocks with language labels."
    }
  },
  "contents": [
    {
      "role": "user",
      "parts": [
        {
          "text": "prompt from user input here"
        }
      ]
    }
  ],
  "generation_config": {
    "temperature": 1,
    "topP": 1
  }
}
```

and using Gemma, the POST looks like: 
```
{
  "contents": [
    {
      "role": "user",
      "parts": [
        {
          "text": "You use markdown liberally to structure responses. Always show code snippets in markdown blocks with language labels."
        }
      ]
    },
    {
      "role": "user",
      "parts": [
        {
          "text": "hello there"
        }
      ]
    }
  ],
  "generation_config": {
    "temperature": 1,
    "topP": 1
  }
}
```

### DIFF
--- a/chatgpt-shell-google.el
+++ b/chatgpt-shell-google.el
@@ -216,6 +216,30 @@ This gets set once for each MODEL, based on a heuristic."
         "google_search_retrieval"
       "google_search")))
 
+(defun chatgpt-shell-google--is-system-instruction-supported (model)
+  "Returns true if system_instruction is supported in the POST payload.
+
+This is based on a heuristic."
+  (when-let* ((current-model model)
+              (is-google (string= (map-elt current-model :provider) "Google"))
+              (version (map-elt current-model :version)))
+    (not (string-match "gemma-3-27b-it" version))))
+
+
+(defun chatgpt-shell-google--maybe-system-prompt (settings model in-sys-instruction)
+  "returns the appropriate portion of prompt depending on SETTINGS, MODEL, and whether
+in system instruction."
+  (let ((sys-prompt (map-elt settings :system-prompt))
+        (sys-instruction-supported
+         (chatgpt-shell-google--is-system-instruction-supported model)))
+    (when sys-prompt
+      (if in-sys-instruction
+          (when sys-instruction-supported
+            `((system_instruction . ((parts . ((text . ,sys-prompt)))))))
+        (when (not sys-instruction-supported)
+          `(((role . "user")
+             (parts . ,(vconcat `(((text . ,sys-prompt))))))))))))
+
 (defun chatgpt-shell-google-models ()
   "Build a list of Google LLM models available."
   ;; Context windows have been verified as of 11/26/2024. See
@@ -319,9 +343,9 @@ or
 
  Compose using PROMPT, CONTEXT, SETTINGS and MODEL."
   (append
-   (when (map-elt settings :system-prompt)
-     `((system_instruction . ((parts . ((text . ,(map-elt settings :system-prompt))))))))
+   (chatgpt-shell-google--maybe-system-prompt settings model 't)
    `((contents . ,(vconcat
+                   (chatgpt-shell-google--maybe-system-prompt settings model 'nil)
                    (chatgpt-shell-google--gemini-user-model-messages
                     (append context
                             (when prompt


xenodium:
Thanks for the PR!

Can't we handle this in `chatgpt-shell-google--validate-command` like we do in `chatgpt-shell-validate-no-system-prompt` and guide the user to unset it? Otherwise, it feels a little misleading. The shell prompt is displaying a system value, but it's not actually used for anything.


image

DinoChiesa:
> The shell prompt is displaying a system value, but it's not actually used for anything.

It is still being used. It's just placed in the user section of the prompt, rather than the system_instruction section of the prompt.  See my example JSON payloads above.

What do you want to do?

BTW This is a bit of an edge-case.  (a) Gemma is new and evolving and it will probably add system_instruction "real soon now."  At which point this change will no longer be necessary.   and (b) most people using chatgpt-shell I suppose won't use Gemma, which is mostly intended for use in embedded scenarios (on the phone, etc) for handling images and videos.  I may be wrong about that; maybe chatgpt-shell is a good tool for handling images in prompts and responses. I have not explored this.

xenodium:
> It is still being used. It's just placed in the user section of the prompt, rather than the system_instruction section of the prompt. See my example JSON payloads above.

Thanks for clarifying. The way I was thinking about it is that if the model doesn't support system prompts, we don't add them to the payload at all, and guide users to unset if needed. I'm a little worried about what potential ambiguity it may cause. Maybe it's not a big deal and I'm overthinking it. Do you by chance know how other projects go about this?

DinoChiesa:
>  Do you by chance know how other projects go about this?

I don't. The space is moving so quickly. I suppose that's why this model doesn't support system instructions; Google's in a hurry. (I don't have any actual insight tho)  And I suppose  Gemma will support system instructions "at some point". 

Maybe the optimal thing is to wait for someone else to scream about it, and if that happens, as you say, "guide users to unset" the behavior.  The problem may just go away. But maybe you have higher aspirations for this project. I mean you may be concerned about users who try it and don't complain, but just .. silently give up.

xenodium:
We can give this a try, though my suggestion would be to inline the changes and apply comments. I found something like the following a little easier to understand why we're applying an exception to the rule:

```emacs-lisp
(cl-defun chatgpt-shell-google--make-gemini-payload (&key prompt context settings model)
  "Create the request payload.

 Compose using PROMPT, CONTEXT, SETTINGS and MODEL."
  (append
   (when (and (map-elt settings :system-prompt)
              ;; Gemma does not support system prompts.
              (not (string-match "gemma-3-27b-it" (map-elt model :version))))
     `((system_instruction . ((parts . ((text . ,(map-elt settings :system-prompt))))))))
   `((contents . ,(vconcat
                   (when (and (map-elt settings :system-prompt)
                              ;; Include system prompt as user message for gemma.
                              (string-match "gemma-3-27b-it" (map-elt model :version)))
                     `(((role . "user")
                        (parts . ,(vconcat `(((text . ,(map-elt settings :system-prompt)))))))))
                   (chatgpt-shell-google--gemini-user-model-messages
                    (append context
                            (when prompt
                              (list (cons prompt nil))))))))
   (when (map-elt model :grounding-search)
     ;; Grounding in Google Search is supported for both Gemini 1.5 and 2.0 models.
     ;; But the API is slightly different between them. This uses the correct tool name.
     `((tools . ((,(intern (chatgpt-shell-google--get-grounding-in-search-tool-keyword model)) . ())))))
   `((generation_config . ((temperature . ,(or (map-elt settings :temperature) 1))
                           ;; 1 is most diverse output.
                           (topP . 1))))))
```

I didn't test the gemma branch. Btw, what model config did you use gemma?

If this works for gemma and you're keen, feel free to apply snippet as part of this PR. Otherwise, I can give it a try later today.

PR #336, opened 2025-03-26
Use extended thinking on Claude 3.7 Sonnet

jukkaaro:
More detailed description of extended thinking in
* https://support.anthropic.com/en/articles/10574485-using-extended-thinking-on-claude-3-7-sonnet
* https://docs.anthropic.com/en/docs/about-claude/models/extended-thinking-models#extended-output-capabilities-beta

### DIFF
--- a/chatgpt-shell-anthropic.el
+++ b/chatgpt-shell-anthropic.el
@@ -95,7 +95,7 @@ VALIDATE-COMMAND handler."
    (chatgpt-shell-anthropic--make-model :version "claude-3-7-sonnet-latest"
                                         :short-version "3.7-sonnet"
                                         :token-width  4
-                                        :max-tokens 8192
+                                        :max-tokens 128000
                                         :context-window 200000)
    (chatgpt-shell-anthropic--make-model :version "claude-3-5-sonnet-latest"
                                         :short-version "3.5-sonnet"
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for the PR. I'll merge and see how we get on with it. We may want to make thinking configurable (ie. at minimum turn on and off).

jukkaaro:
> Thanks for the PR. I'll merge and see how we get on with it. We may want to make thinking configurable (ie. at minimum turn on and off) in the future.

Thanks @xenodium , I was also thinking about making this configurable but did not have time to do that yet. I assume switching extended thinking on/off is actually required, otherwise one would e. g. spend more tokens than necessary for certain requests where extended thinking is not really required.

I'll see if I find time to add the option for on/off and will post it / open another PR later.

xenodium:
We don't yet have a centralized way of dealing with settings that cater for model differences. The generic mechanism is a bigger change and needs more thought.

I think for now, we can have a model-specific buffer-local variable. I can maybe add that shortly.

xenodium:
Added a toggle in https://github.com/xenodium/chatgpt-shell/commit/23ebad074e1c0b3d6bd2d8663dd149103e71911f Value is off by default. Give that a try.

djr7C4:
> We don't yet have a centralized way of dealing with settings that cater for model differences. The generic mechanism is a bigger change and needs more thought.
> 
> I think for now, we can have a model-specific buffer-local variable. I can maybe add that shortly.

I've been thinking about this a bit. Just because we add a UI doesn't mean these variables need to go away. We can just add a field to the model that has a list of the variables that the user should be able to configure. 

It would also contain information on how to choose new values for them. The easiest way would probably be to just provide a function that could be called.

Some values will need to be read mostly free form for things like temperature. We could use competing-read for variables that only take certain values from a list. An alternative would be a transient menu with all the configuration options. It would need to trigger read for variables with continuous values.

PR #337, opened 2025-03-27
add gemini-2.5-pro-exp

whalliburton:
New google model.

### DIFF
--- a/chatgpt-shell-google.el
+++ b/chatgpt-shell-google.el
@@ -221,7 +221,12 @@ This gets set once for each MODEL, based on a heuristic."
   "Build a list of Google LLM models available."
   ;; Context windows have been verified as of 11/26/2024. See
   ;; https://ai.google.dev/gemini-api/docs/models/gemini.
-  (list (chatgpt-shell-google-make-model :version "gemini-2.0-pro-exp"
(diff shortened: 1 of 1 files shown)


djr7C4:
According to the documentation, the context window should be 1000000 instead of 2048576.

https://ai.google.dev/gemini-api/docs/models

DinoChiesa:
If you use `chatgpt-shell-google-load-models`, you will get the definition for the new gemini-2.5-pro-exp model.  And then if you use `chatgpt-shell-google-load-models`, you will be able to select it.   I have these commands bound to `C-c l` and `C-c C-v` respectively in my emacs.

Loaded this way, gemini-2.5-pro-exp-03-25  uses a max context window of 1048576, which is what Google's API says it should be.  It seems like the documentation does not agree with the API.

xenodium (approved):

DinoChiesa:
The context window documented [here](https://ai.google.dev/gemini-api/docs/models#gemini-2.5-pro-exp-03-25) for gemini-2.5-pro-exp-03-25 is 1048576.    
In contrast, the context window size in this PR is **2**048576. 

image

djr7C4:
xenodium already fixed it when this was merged.

DinoChiesa:
Thanks, I had missed that.

PR #346, opened 2025-04-15
Add Gemini 2.5 pro preview

djr7C4:
This is the same model as 2.5 pro experimental. The difference is that it has higher rate limits but is also paid.


xenodium (approved):
I don't even know what to use anymore! Thanks for keeping up!

PR #352, opened 2025-05-10
Transform chatgpt-shell-proofread-region to a DWIM approach

christophe-gouel:
Thanks for your package.

Here is a small tweak to `chatgpt-shell-proofread-region` to decrease friction: if no region is active, mark paragraph and proofread this paragraph. Since this function is likely to be run paragraph by paragraph, this removes the need for selecting a paragraph.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -1469,11 +1469,22 @@ If region is active, append to prompt."
 
 ;;;###autoload
 (defun chatgpt-shell-proofread-region ()
-  "Proofread text from region using ChatGPT.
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thank you!

xenodium:
We may want to follow-up renaming the function. Some options:

- chatgpt-shell-proofread
- chatgpt-shell-paragraph-or-region

Maybe alias chatgpt-shell-proofread-region to the new function to avoid breaking users.

Happy to do this, unless you want to.

PR #353, opened 2025-05-10
Rename proofread function, and add alias for old name

christophe-gouel:
You are right. It is useful to rename the function. Here is the PR.

Thanks for merging so quickly.

Note that there is one tiny default to what I proposed. When proofreading a paragraph, if the AI suggestion is rejected, instead of the point staying at its original position, the point is moved to the beginning of the paragraph and the paragraph is marked. I have tried a few things, but I was not able to solve this, but this is a minor issue that I can live with.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -189,6 +189,8 @@ For example:
 
 (defalias 'chatgpt-shell-save-session-transcript #'shell-maker-save-session-transcript)
 
+(defalias 'chatgpt-shell-proofread-region #'chatgpt-shell-proofread-paragraph-or-region)
+
 (defvar chatgpt-shell--prompt-history nil)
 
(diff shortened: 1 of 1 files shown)


xenodium (approved):

xenodium:
Thank you!

xenodium:
> Note that there is one tiny default to what I proposed. When proofreading a paragraph, if the AI suggestion is rejected, instead of the point staying at its original position, the point is moved to the beginning of the paragraph and the paragraph is marked. I have tried a few things, but I was not able to solve this, but this is a minor issue that I can live with.

Ah yes. I've been meaning to look into it... What happens is that even if we reject the change, the buffer was modified anyway (so smerge could display the diff). If keen, maybe take a look at `chatgpt-shell--pretty-smerge-insert`. It may be possible leverage undo stack?

PR #361, opened 2025-05-24
Update the Google modes

ccqpein:


ccqpein:
> This replaces 2.5-pro with flash. That's not an update. It's replacing a more powerful model with a faster one. IMO 2.5 flash should be added but as a separate model.

Yes, I remove it because I don't find the [free tier](https://ai.google.dev/gemini-api/docs/rate-limits#current-rate-limits) of that model, so it might cost default. I can add it back, but it might push some risk to people who use it but are not expecting it will cost some money.

> what does grounding-search do and do we want it enabled by default?

This will let Gemini use Google’s search as the input knowledge for giving the real-time results. Ref #303

djr7C4:
I use 2.5 pro and wouldn't like to see it removed. IMO people shouldn't use the API if they are unaware of how billing works. Charges are the expectation.

ccqpein:
> Charges are the expectation

emmmm, I think you are right. I just put it back.

PR #368, opened 2025-06-13
Allow the maximum number of thinking tokens to be allowed for Claude

djr7C4:
By default chatgpt-shell-anthropic-thinking-budget-tokens is 32000. This exceeds
the maximum number of tokens for some models (e.g. Opus 4) and is a pain since
the maximum number of tokens differs across different anthropic models. This PR
allows a value of 'max which uses as many thinking tokens as possible.


xenodium (approved):
Thanks for the PR and the follow-up change.

PR #369, opened 2025-06-18
Make the screenshot program configurable

djr7C4:


chatgpt-shell.el
```diff
@@ -2025,6 +2025,18 @@ If in a `dired' buffer, use selection (single image only for now)."
                           :prompt-url file
                           :streaming nil)))
 
+(defcustom chatgpt-shell-screenshot-program
```
xenodium, line 2028: We can likely consolidate a little if we rename this to something like chatgpt-shell-screenshot-command, making it a list which includes both the program and all its arguments. Also reduces the chance of mismatch between program and associated args, since they are both in the same place.

xenodium (commented):
Thanks for this! I'm surprised I hadn't heard about this change before :)

Did this post, by chance, lead you here? https://xenodium.com/a-platform-that-moulds-to-your-needs

In any case, configuring sounds great. Mind if we consolidate? See comment

djr7C4:
Yes, I did come across it from that post! I'll make the change.

xenodium:
We needed a few more tweaks:

https://github.com/xenodium/chatgpt-shell/commit/c7b7b207f3b6780479fd4af2683a19fcb09ee3e8
https://github.com/xenodium/chatgpt-shell/commit/5e2cc001e1129456363d2c55f7e336c9f00aac79

xenodium:
Also, doesn't the imagemagick import command need additional params to take a screenshot and save to a file? While I haven't tested on Linux, just running `import file/to/output.png` doesn't seem to cut it.

xenodium:
Looks like it's a macOS thing so we're good on Linux, which you tested anyway. Thanks for following up.

PR #370, opened 2025-06-20
minor: Update Gemini version

ccqpein:


djr7C4:
Nice! I see that 2.5 is out now. I'd add 2.5 pro as well since it has also been released.

xenodium (approved):
Thank you!

PR #376, opened 2025-07-20
fix: Foreign buffer fontification (fixes #375)

### LINKED ISSUE #375: chatgpt-shell fontifies wrong buffer when shell buffer not current on completion
To reproduce (I suppose streaming must be enabled):

1. Create arbitrary buffer with markdown syntax `*abuf*`
2. Invoke `chatgpt-shell-prompt` and send some prompt
3. While the compose buffer is completing the response switch back to the original buffer `*abuf*`
4. When the response finished chatgpt-shell will "fontify" `*abuf*` with markdown overlays

This is can be quite a problem with programming language buffers since their syntax sometimes doubles as markdown, e.g.
```go
func makeFeed(feed *feed.Feed, cache *cache.Cache)
```
After chatgpt is finished it will italicize *feed.Feed, cache*

arminfriedl:
Fix fontification of arbitrary buffers. Currently, with streaming enabled, it is possible to switch away from the chatgpt-shell buffer while it is finishing writing the response. Once this happened the `on-command-finished` puts markdown overlays in the current buffer, which might not be the chatgpt-shell buffer.

Wrap markdown overlay fontification in
`(with-current-buffer (chatgpt-shell--primary-buffer))` s.t. it puts the overlays in the `chatgpt-shell` buffer.

### DIFF
--- a/chatgpt-shell.el
+++ b/chatgpt-shell.el
@@ -707,7 +707,8 @@ See `shell-maker-welcome-message' as an example."
        (error "%s not found" (chatgpt-shell-model-version))))
    :on-command-finished
    (lambda (command output success)
-     (markdown-overlays-put)
+     (with-current-buffer (chatgpt-shell--primary-buffer)
+       (markdown-overlays-put))
      (run-hook-with-args 'chatgpt-shell-after-command-functions
                          command output success))
    :redact-log-output


xenodium:
Thank you for filing a bug and sending a PR!

I'm thinking if we fix in shell-maker, we make it more generic and avoid depending on primary shells to address the issue (a little rigid and coincidentally I've considering modifying the concept of primary buffers).

I've submitted a fix in shell-maker https://github.com/xenodium/shell-maker/commit/d65fbc76ccc4c8e0bf77b62b492ff05710d0a4d5

Mind updating shell-buffer to v0.78.2 and seeing if that resolves the issue for ya?

Issue #374, opened 2025-07-17
OpenRouter is cutting off after a few tokens


rogsme:
As titled. I'm using `anthropic/claude-3.7-sonnet` and the output is very inconsistent

<img width="722" height="592" alt="Image" src="https://github.com/user-attachments/assets/8c026f39-ed73-4762-9358-f5609241055c" />

I tried to add `google/gemini-2.5-flash` and it has the same issues, even worse:

<img width="721" height="417" alt="Image" src="https://github.com/user-attachments/assets/c59ba08b-6ed3-4072-b849-a69a99be22a9" />

To give more context, here are the OpenRouter logs for the 3 Claude calls and the 3 Gemini-2.5-flash calls:

<img width="1125" height="392" alt="Image" src="https://github.com/user-attachments/assets/a9ab8806-2229-42ca-9644-d7ef6c8a701b" />

They were all served by Google Vertex. Maybe there's a bug on how Google Vertex serves the responses?

unholywhale:
I'm getting the same problem - any model I tested on OpenRouter is frequently dropping tokens

xenodium:
Alternatively, @unholywhale @rogsme investigating and patching super welcome ;) [chatgpt-shell-openrouter.el](https://github.com/xenodium/chatgpt-shell/blob/main/chatgpt-shell-openrouter.el) is a good start.

Logging can be useful too. `(setq chatgpt-shell-logging t)` and look for the *chatgpt-log* buffer.

djr7C4:
I can reproduce for claude 3.7 sonnet but it seems to work fine for some others (e.g. qwen-2.5-coder-32b). It's not immediately clear to me why it stopped working. OpenRouter, OpenAI and Anthropic all have the same API so there is something weird going on here...

xenodium:
> I can reproduce for claude 3.7 sonnet

Via `chatgpt-shell-anthropic.el` or `chatgpt-shell-openrouter.el`?

rogsme:
PR was merged, I'll mark this issue as closed

PR #378, opened 2025-07-25
Fix openrouter filter output function

unholywhale:
Problem:
Tokens are being dropped when using OpenRouter models #374

Root cause:
Fragmented JSON chunks and ": OPENROUTER PROCESSING" spam in the streaming output.

Solution:
Process streaming output line-by-line with a temporary buffer
Concat received ":pending" with new output
Drop lines that start with ":" - "OPENROUTER..." spam
Only process complete "data: {...}"
Incomplete lines are passed as :pending for the next cycle

### DIFF
--- a/chatgpt-shell-openrouter.el
+++ b/chatgpt-shell-openrouter.el
@@ -164,7 +164,6 @@ If you use OpenRouter through a proxy service, change the URL base."
    :filter #'chatgpt-shell-openrouter--filter-output
    :missing-key-msg "Your chatgpt-shell-openrouter-key is missing"))
 
-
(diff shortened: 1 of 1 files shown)


rogsme:
I can confirm this fix works! 

image

image

Great work @unholywhale 🔥

rogsme (approved):

xenodium:
@unholywhale Thanks for looking into it and fixing! LGTM. @djr7C4 fancy reviewing also?

djr7C4 (approved):

djr7C4:
Nice.

PR #381, opened 2025-07-30
Add correct transient dependency

rogsme:
## Description
This PR addresses a lazy loading issue with the `transient` package. Previously, the `chatgpt-shell-transient` functionality, which defines the transient popup menu for `chatgpt-shell`, was not wrapped in `with-eval-after-load`. This could lead to errors if `transient` was not fully loaded before `chatgpt-shell-transient.el` was evaluated.

The fix involves encapsulating the `transient-define-prefix` and `transient-define-suffix` calls within an `(with-eval-after-load 'transient ...)` block. This ensures that the transient definitions are only evaluated after the `transient` package has been loaded, preventing potential initialization errors.

Additionally, the `Package-Requires` in `chatgpt-shell.el` has been updated to explicitly include `transient` with a minimum version of "0.9.3", reflecting its direct dependency.

## How Has This Been Tested?
The changes were manually tested by:
- Loading `chatgpt-shell` and verifying that the `chatgpt-shell-transient` popup correctly appears and functions as expected.
- Ensuring no errors related to `transient` were thrown during Emacs startup or package loading.
- Confirming that all commands accessible through the transient menu still work correctly.

### DIFF
--- a/chatgpt-shell-transient.el
+++ b/chatgpt-shell-transient.el
@@ -7,97 +7,96 @@
 
 (require 'transient)
 
-(defun chatgpt-shell-transient--in-shell-p ()
-  "Return non-nil if the current buffer is in chatgpt-shell-mode."
-  (derived-mode-p 'chatgpt-shell-mode))
+(with-eval-after-load 'transient
+  (defun chatgpt-shell-transient--in-shell-p ()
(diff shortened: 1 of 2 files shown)


xenodium:
Thank you!

PR #385, opened 2025-08-07
Add GPT 5 support

djr7C4:


djr7C4:
Huh. I didn't have to submit anything new. I've used it in some scripts and
am on tier 2 so I'm not sure if that makes a difference. I had to submit a
driver's license at one point to get access to o1 IIRC. Is that what you
mean?

xenodium:
>  I had to submit a driver's license at one point to get access to o1 IIRC. Is that what you
mean?

Ah interesting. I never had to. Even for o1. I may have been grandfathered into it, but is now required for some models.

PR #388, opened 2025-08-09
Support setting the reasoning effort interactively and add opus 4.1

djr7C4:


xenodium:
Thanks for this. Mind separatng and only adding "opus-4-1-20250805" in this PR? We can treat the thinking logic separately.

xenodium:
Can we do it maybe the other way around? Rename this PR to be about the thinking budget and open a new with the model addition only?

```diff
+   (chatgpt-shell-anthropic--make-model :version "claude-opus-4-1-20250805"
+                                        :short-version "opus-4.1"
+                                        :token-width  4
...)
```

Alternatively, we can rename the PR to state all the functional changes (not just adding opus 4.1).

Btw, have you run `M-x checkdoc` and byte compiled the files?

djr7C4:
Done. I went with option 2. I haven't run checkdoc yet but I can tomorrow.

djr7C4:
Checkdoc issues should be fixed now.

chatgpt-shell.el
```diff
@@ -608,6 +608,53 @@ non-nil; otherwise `completing-read'."
...
+      (user-error "No reasoning effort selector is defined for %s" (chatgpt-shell-model-version)))
+    (let* ((first t)
+           (buf (cond
+                 (global
+                  nil)
+                 ((eq major-mode 'chatgpt-shell-mode)
+                  (current-buffer))
+                 ((memq major-mode '(chatgpt-shell-prompt-compose-mode chatgpt-shell-prompt-compose-view-mode))
+                  (chatgpt-shell--primary-buffer))))
+           (result (with-current-buffer buf (funcall selector model))))
```
xenodium, line 645: Can we give this var a name a little less generic? If I'm understanding correctly, it's got details about the variable that should be set?

xenodium, line 645: Can we add comment showing what the output of `(funcall selector model)` can look like? Without inspecting the output at runtime, it's hard to know what to expect.

chatgpt-shell.el
```diff
@@ -608,6 +608,53 @@ non-nil; otherwise `completing-read'."
...
+      (dolist (cell result)
```
xenodium, line 646: Curious, what's `cell` referring to? It's not immediately obvious from its name and context. Both cell and result are fairly generic. Can we name differently? What are we iterating over?

chatgpt-shell.el
```diff
@@ -608,6 +608,53 @@ non-nil; otherwise `completing-read'."
...
+        (let* ((sym (car cell))
+               (var (if buf
+                        (with-current-buffer buf
+                          (make-local-variable sym))
+                      sym))
+               (val (cdr cell)))
+          (set var val)
+          (when first
```
xenodium, line 654: It's not super obvious why we're showing a message on the first iteration.

chatgpt-shell.el
```diff
@@ -608,6 +608,53 @@ non-nil; otherwise `completing-read'."
...
+            (message "Set %s to %s%s" var (if val val "max") (if buf " locally" " globally"))
```
xenodium, line 655: If there's no val, isn't it technically unsetting the variable? Stating it's set to max could be misleading, specially when troubleshooting something that isn't working.

chatgpt-shell-openai.el
```diff
@@ -46,7 +47,19 @@ for more difficult problems."
...
   :group 'chatgpt-shell)
 
-(cl-defun chatgpt-shell-openai-make-model (&key version short-version token-width context-window validate-command (headers #'chatgpt-shell-openai--make-headers) (key chatgpt-shell-openai-key) (url-base 'chatgpt-shell-api-url-base) (path "/v1/chat/completions") (provider "OpenAI") (label "ChatGPT") (handler #'chatgpt-shell-openai--handle-chatgpt-command) (filter #'chatgpt-shell-openai--filter-output) reasoning-effort icon function-calling other-params)
+(defun chatgpt-shell-openai-make-reasoning-effort-selector (choices)
+  "Create a function for the :reasoning-effort-selector parameter using CHOICES."
+  (lambda (_model)
+    (let ((effort (completing-read "Reasoning effort: " (chatgpt-shell-unsorted-collection choices) nil t)))
+      `((chatgpt-shell-openai-reasoning-effort . ,effort)))))
+
+(defalias 'chatgpt-shell-openai-old-reasoning-effort-selector
```
xenodium, line 56: Seeing "old" kinda threw me off a little. Like is it deprecated? something else? Seeing its usage in "gpt-4.1" clarified. If I'm understanding correctly, the valid values are just different across some models, correct?

You added a wonderful helper already, I think we can do without aliases and just use the helpers directly in the model definitions:

:reasoning-effort-selector . ,(chatgpt-shell-openai-make-reasoning-effort-selector '("low" "medium" "high"))

chatgpt-shell.el
```diff
@@ -608,6 +608,53 @@ non-nil; otherwise `completing-read'."
...
+                  (chatgpt-shell--primary-buffer))))
+           (result (with-current-buffer buf (funcall selector model))))
+      (dolist (cell result)
+        (let* ((sym (car cell))
+               (var (if buf
+                        (with-current-buffer buf
+                          (make-local-variable sym))
+                      sym))
+               (val (cdr cell)))
+          (set var val)
```
xenodium, line 653: I'm guessing this is creating a variable that can be looked up using `describe-variable`? Where is the variables declared?

chatgpt-shell.el
```diff
@@ -608,6 +608,53 @@ non-nil; otherwise `completing-read'."
         (setq-default chatgpt-shell-model-version selection))
     (error "No other providers found")))
 
+(defun chatgpt-shell-unsorted-collection (collection)
```
xenodium, line 611: For my understanding, why do we need this? For me, `(completing-read "this: " '(B A C))` seems to respect the order.

ps. If we need to keep this, we likely want it to be a private function (ie. chatgpt-shell-- prefix)

xenodium (changes requested):

chatgpt-shell.el
djr7C4: Changed to bindings

chatgpt-shell.el
djr7C4: It refers to a cons cell in the alist. I'll rename it to binding.

chatgpt-shell.el
djr7C4: No, nil values mean the max for thinking budgets. I added a comment.

chatgpt-shell-openai.el
djr7C4: I meant it's the older convention where minimal is not supported. I'll remove the aliases.

chatgpt-shell.el
djr7C4: No, these variables are all already declared. Calling `make-local-variable` is a standard method used in elisp code to create a value that causes it to be set buffer locally when `set` is used. I'll add a comment.

chatgpt-shell.el
djr7C4: This depends on the completion framework that you use. Some completion frameworks sort the candidates based on various factors (e.g. alphabetical order and historical choices). Depending on which you use, it might change the order if you call it multiple times and make different selections.

I'll make it private.

djr7C4:
Thanks for the comments. They should be addressed now.

chatgpt-shell.el
```diff
@@ -608,6 +608,70 @@ non-nil; otherwise `completing-read'."
...
+    (let* ((first t)
+           (buf (cond
+                 (global
+                  nil)
+                 ((eq major-mode 'chatgpt-shell-mode)
+                  (current-buffer))
+                 ((memq major-mode '(chatgpt-shell-prompt-compose-mode chatgpt-shell-prompt-compose-view-mode))
+                  (chatgpt-shell--primary-buffer))))
+           (bindings (with-current-buffer buf
+                       ;; This call returns an alist that maps variables to the
```
xenodium, line 648: Thanks for the comment. Sorry, I should have been more explicit. Could you show the actual structure in an exampe. Something like:

````emacs-lisp
`((foo . "hello")
  (bar . "bye"))
````

It's easier to parse actual data than piece together the comment.

File: chatgpt-shell.el (hunk 6 above)
Comment (xenodium) [reply chatgpt-shell.el:662]: Thanks for this.

> No, these variables are all already declared

I missed them. Could you share the name of one as an example, so I can search for them? I would also add this as a comment in code.

File: chatgpt-shell.el (hunk 3 above)
Comment (xenodium) [reply chatgpt-shell.el:676]: Thanks for clarifying.

Relying on the first item of a list to mean something is a little fragile. Can we make this contract a bit more explicit? Can't we use an alist for each binding to be more descriptive? For example:

```emacs-lisp
((:symbol . chatgpt-shell-anthropic-thinking-budget-tokens)
 (:value . 3000)
 (:kind . thinking-budget))
````

Coincidentaly, this would make the client code in this function much more self-descriptive (and concise) as we drop the `first`variable, only need a single let `binding` var and everything else can use `(map-elt binding :symbol)`, `(map-elt binding :vaue)` and `(map-elt binding :kind)`.

Also, no implicit checks.

```emacs-lisp
(when first
  ...)
```

 becomes

```emacs-lisp
(when (eq map-elt binding :kind)
                 'thinking-budget)
  ...)
```

chatgpt-shell.el
xenodium: You're right. I missed the vars in the example. Super handy to see them in the comment as anyone can read the comment and know what to search for to find related code. Similarly, using an alist as proposed in my previous comment also strengthens this searchable aspect.

chatgpt-shell.el
xenodium: Ah, I see. Did you need it for your completion framework? Which one are you using?

djr7C4: Yes, for mine  (vertico) the order can change by default.

djr7C4:
Ok, I reworked the structure of the bindings.

xenodium:
Thank you. I appreciate the changes. Almost there. Can we complete the returned binding so it has a consistent structure like:

```emacs-lisp
((:symbol . chatgpt-shell-anthropic-thinking-budget-tokens)
 (:value . 3000)
 (:kind . thinking-budget))
```

At the moment, we store symbol and vaue one way but kind differently.

djr7C4:
I dislike adding extra keys like this for arguments that are required anyway. It adds unnecessary verbosity and allows required arguments to be omitted. As for storing things differently, it is no different from parameters to a function defined with cl-defun since these work the same way as a destructuring-bind.

xenodium:
> It adds unnecessary verbosity

Sometimes warranted if it simplifies maintenance. In this case, it makes the returned schema heterogeneous an thus easier to visualize.

In addition, it makes the code super explicit. When we read `(map-elt binding :kind)`, we know exactly what it means (the binding kind) without having to read anywhere else in the function. Requiring additional local variables add another level of indirection.

> to a function defined with cl-defun since these work the same way as a destructuring-bind.

We don't use destructuring-bind anywhere else in the codebase. `seq-elt` is fairly prevalent now. It'd be preferable to maintain some consistency. While I understand it's a personal choice to use `seq-elt` vs `destructuring-bind`, as maintainer, I have to try to steer to adopt some idioms. Including everyone's preferred way of doing things would unfortunatey yield code that's harder to maintain over time. Hope that makes sense.

Btw, it's not a huge change from what you already have:

```emacs-lisp
(defun chatgpt-shell-select-reasoning-effort (&optional global)
  "Interactively set the reasoning effort for the current model.

By default, this is done buffer-locally when in a
`chatgpt-shell-mode' buffer `chatgpt-shell-prompt-compose-mode'
buffer. When GLOBAL is non-nil (interactively with a prefix
argument), it is set globally."
  (interactive "P")
  (let* ((model (chatgpt-shell--resolved-model))
         (selector (map-elt model :reasoning-effort-selector)))
    (unless selector
      (user-error "No reasoning effort selector is defined for %s" (chatgpt-shell-model-version)))
    (let* ((buf (cond
                 (global
                  nil)
                 ((eq major-mode 'chatgpt-shell-mode)
                  (current-buffer))
                 ((memq major-mode '(chatgpt-shell-prompt-compose-mode chatgpt-shell-prompt-compose-view-mode))
                  (chatgpt-shell--primary-buffer))))
           ;; The call to the selector returns a list of bindings. Some models
           ;; (e.g. those by Anthropic) have multiple variables that control
           ;; reasoning so in some cases it is necessary to set more than one.
           ;; An example return value is
           ;;
           ;; ((chatgpt-shell-anthropic-thinking-budget-tokens 3000 :kind thinking-budget)
           ;;  (chatgpt-shell-anthropic-thinking t :kind thinking-toggle))
           (bindings (if buf
                         (with-current-buffer buf
                           (funcall selector model))
                       (funcall selector model))))
      (dolist (binding bindings)
        (unless (memq (map-elt binding :kind)
                      '(thinking-budget thinking-toggle))
          (error "Unknown kind %S returned by reasoning effort selector" (map-elt binding :kind)))
        (if buf
            ;; Ensure that the variable is set buffer-locally.
            (with-current-buffer buf
              (set (make-local-variable (map-elt binding :symbol))
                   (map-elt binding :value)))
          ;; Set the global value even if it has already been bound
          ;; buffer-locally. Note that using `set' on var will set the
          ;; buffer-local value if one already exists.
          (set-default var val))
        ;; Let the user know what the thinking budget was set to.
        (when (eq (map-elt binding :kind) 'thinking-budget)
          (message "Set %s to %s%s"
                   (map-elt binding :symbol)
                   (if max "max" (map-elt binding :value))
                   (if buf " locally" " globally")))))))
```

djr7C4:
I don't agree but it's your perogative as maintainer. I've made the changes.

djr7C4:
BTW, I just want to say that the level of changes required for this PR was a lot more than I expected. I spent a lot more time meeting all the requirements than I did on the initial code.

I like contributing my changes back to the project but from a time efficiency perspective it might make more sense for me to just modify my own configuration or personal fork for cases like this. I'd estimate this process took at least three times longer than it would have if I hadn't submitted the PR.

I understand your rationale and there are no hard feelings but I will have to think more carefully next time about the additional time that may be required when considering making future non-trivial PRs. It's your perogative as maintainer to require whatever changes you feel are best. However, as a contributor I'm going to be less likely to submit PRs if it takes this much extra time.

xenodium (approved):

djr7C4:
Thanks! Yeah, discussing it before might have been a good idea I suppose. I
was thinking more like a brief list of style guidelines. I can think of
several based on this PR.

PR #393, opened 2025-08-16
Support Ollama’s “MXFP*” quantization names when parsing token width

kromancer:
`chatgpt-shell-ollama--parse-token-width` assumed that every quantization
string began with either ‘F’ or ‘Q’.  
Models served by Ollama (e.g. `gpt-oss`) can instead return names such
as “MXFP4”.

### DIFF
--- a/chatgpt-shell-ollama.el
+++ b/chatgpt-shell-ollama.el
@@ -115,7 +115,8 @@ If you use Ollama through a proxy service, change the URL base."
 
 (defun chatgpt-shell-ollama--parse-token-width (quantization)
   "Parse token using QUANTIZATION."
-  (when (string-match "^[FQ]\\([1-9][0-9]*\\)" quantization)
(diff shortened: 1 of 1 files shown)


chatgpt-shell-ollama.el
```diff
@@ -115,7 +115,7 @@ If you use Ollama through a proxy service, change the URL base."
 
 (defun chatgpt-shell-ollama--parse-token-width (quantization)
   "Parse token using QUANTIZATION."
-  (when (string-match "^[FQ]\\([1-9][0-9]*\\)" quantization)
+  (when (string-match "^[A-Z]+\\([1-9][0-9]*\\)" quantization)
```
xenodium, line 118: Minding pasting a sample matched string above the when? Helps the next person visualize actual content coming from LLMs.

xenodium (approved):
Thanks for the PR! Minor request added.

Issue #394, opened 2025-08-21
in chatgpt-shell-google, error: wrong-type-argument consp nil.   The logic for thinkingConfig is not quite correct .


DinoChiesa:
The cl-defun `chatgpt-shell-google--make-gemini-payload` uses a conditional within the `generation_config` param to maybe add information about `thinkingConfig` . But when the result of the condition is false this logic adds a nil to the `generation_config` list. The items in the list are expected to be cons cells. This breaks the attempted subsequent json serialization in shell-maker. 

<details><summary>Stacktrace</summary>
<p>
You can see that the `:data` field ends with `(generation_config (temperature . 1) (topP . 1) nil)` .  It is this final `nil` that is causing the problem. 

</p>
```
Debugger entered--Lisp error: (wrong-type-argument consp nil)
  shell-maker--json-encode(((system_instruction (parts (text . "You use markdown liberally to structure responses. Always show code snippets in markdown blocks with language labels."))) (contents . [((role . "user") (parts . [(...)]))]) (tools (google_search)) (generation_config (temperature . 1) (topP . 1) nil)))
  shell-maker-make--curl-command(:url "https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-pro:streamGenerateContent?key=AI..." :data ((system_instruction (parts (text . "You use markdown liberally to structure responses. Always show code snippets in markdown blocks with language labels."))) (contents . [((role . "user") (parts . [(...)]))]) (tools (google_search)) (generation_config (temperature . 1) (topP . 1) nil)) :encoding nil :timeout nil :headers ("Content-Type: application/json; charset=utf-8") :fields nil :proxy nil)
[...]

PR #395, opened 2025-08-21
fix: if thinking is not available don't add an item

DinoChiesa:
fixes issue #394

### DIFF
--- a/chatgpt-shell-google.el
+++ b/chatgpt-shell-google.el
@@ -374,23 +374,23 @@ or
                            (topP . 1)
                            ;; Include thinking parameters if it is supported for
                            ;; this model.
-                           ,(let ((min (map-elt model :thinking-budget-min))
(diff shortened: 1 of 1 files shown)


chatgpt-shell-google.el
```diff
@@ -174,22 +174,46 @@ https://generativelanguage.googleapis.com"
                                        :token-width 4
                                        :context-window .inputTokenLimit))))
 
+(defun chatgpt-shell-google-use-default-models ()
```
xenodium, line 177: Can't we use `M-x chatgpt-shell-reload-default-models`?

xenodium (commented):
Thanks for the PR! I've not had a chance to apply and try, but left a question for now.

chatgpt-shell-google.el
DinoChiesa: YES

xenodium:
Thank you. Merged.

PR #400, opened 2025-09-22
Enable function calling for openai models that support it

djr7C4:


djr7C4:
I've verified that these models support function calling according to OpenAI's website

xenodium:
Thank you. I gotta say, I haven't used function calling much since I added this.

Is it working well for you? What are you using it for?

Btw, I'm working on an agent shell now which delegates all this stuff to the agent. It's early days still, so not yet public but happy to add you to the repo if you'd like to take it for a spin (only Claude Code and Gemini CLI support at the moment).

djr7C4:
Sure. I'd be interested in taking a look at it. Thanks! So far I've only tested tool use with your weather tool but it worked as intended.

I'm interested in writing a package that allows the user to modify code in the current buffer by voice. Think less along lines of something like Aider and more like "add a class with slots named ..." or "add the slots ... too the class ..." At the moment I'm thinking about how tool use and agents factor into this.

I'm unsure if it will make sense to base it on chatgpt-shell or something else. In addition to tool use for other providers, I'd need to add audio input capabilities or use something like whisper.el. It might be faster to use a model that supports audio input rather than transcribing separately though.

I also looked into llm.el but it seems to abstract away too much from the models in order to provide a consistent interface.

I've tested some models. OpenAI tends to be slow. Google is fairly fast but still not as fast as would be ideal for audio. gpt-20b is very fast when you go with the lowest latency provider but it doesn't support audio.

djr7C4:
Thanks. I'll check it out. I'm on Linux BTW.

PR #415, opened 2025-12-17
Add the 3.0 flash preview

ccqpein:
Google just release the 3.0 flash this morning. Add it in this PR

**Some changes**

According to this ([link](https://ai.google.dev/gemini-api/docs/gemini-3)): 

>Important: You cannot use both thinking_level and the legacy thinking_budget parameter in the same request. Doing so will return a 400 error.

I removed the thinking_budget in 3.0 models. We may need the `thinking_level` for 3.0 models in future.


xenodium (approved):

djr7C4:
Should the thinking budget really be removed? The docs say that it is still supported even though the new thinking level system seems to be preferred.

PR #421, opened 2026-02-21
Gemini add 3.1 pro preview

ccqpein:
[doc](https://ai.google.dev/gemini-api/docs/models/gemini-3.1-pro-preview)


xenodium (approved):
Thank you!

PR #425, opened 2026-05-01
Add Mistral provider

jro:
Adds chatgpt-shell-mistral.el with support for Mistral's La Plateforme API (api.mistral.ai), following the same OpenAI-compatible pattern used by the DeepSeek provider. Created with the help of Claude, especially the bits for the Magistral reasoning models.                                                                                          
                                                                                                                                      
  Models included:                                                                                                                    
  - mistral-large-latest, mistral-medium-latest, mistral-small-latest                                                                 
  - codestral-latest (256k context), devstral-latest                                                                                  
  - magistral-medium-latest, magistral-small-latest (reasoning models)                                                                
  - open-mistral-nemo, ministral-14b-latest, ministral-8b-latest, ministral-3b-latest                                                 
                                                                                                                                      
  Magistral reasoning model support:                                                                                                  
                                                                                                                                      
  * Magistral models stream content as an array of typed blocks rather than a plain string:              
  ```                                                                                                                    
  [{"type": "thinking", "thinking": [...]}, {"type": "text", "text": "..."}]                                                       
  ```                                                                                                                                    
  *  The provider includes a custom filter (chatgpt-shell-mistral--filter-output) and helper (chatgpt-shell-mistral--content-text) that handles both the standard string format used by all other Mistral models and this array format. Thinking blocks are discarded; only text blocks are surfaced.


xenodium:
Thank you for the PR and sorry for the delay. Just a heads-up... I'm hardly ever using chatgpt-shell these days. agent-shell has replaced all my previous flows. I use temp agent-shell buffers like I used to use chatgpt-shell

image

jro:
> Thank you for the PR and sorry for the delay. Just a heads-up... I'm hardly ever using chatgpt-shell these days. agent-shell has replaced all my previous flows. I use temp agent-shell buffers like I used to use chatgpt-shell

Nice! I hadn't updated agent-shell in a while, so I've been using an empty directory for something similar. I think I was initially using chatgpt-shell to try and understand the differences between which models were available via agent apis and which were available via chat apis.

PR #430, opened 2026-09-18
Add Perplexity integration attribution

qirh:
## Summary

Perplexity would like to identify requests made through chatgpt-shell. This adds `X-Pplx-Integration: chatgpt-shell` to the existing Perplexity request headers without changing authentication, payloads, or routing.

## Verification

The Perplexity header builder now returns the integration header alongside its existing content type and authorization headers. No other provider code is changed.


xenodium (approved):
Thank you

